PR #14, opened 2024-03-26
Starts work on first integration test

schuylermartin45:
This first-pass of an integration test runs a bulk conversion operation against a few recipe files. In the future, I anticipate adding many more recipe files from various sources, to get a clear picture at how the conversion process is going.

In essence, the new integration test...
- Installs the project
- Converts all recipes in the integration test folder (1 instance per category/directory of files)
- Takes the converted recipes and performs a dry-run build using rattler-build of each recipe that has been converted.
  - This does not exclude files that had errors in the conversion process. If a `recipe.yaml` exists in the test directory, it will undergo a dry-run build.
- The conversion and dry-run-build phases must pass a minimum success threshold in order to "pass"

Some notes:
- Success in the conversion phase is measured by meeting a threshold % of "successful" conversions
- Success in the rattler-build phase is measured by how many builds return 0

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -109,9 +109,10 @@ jobs:
               ${{ runner.os }}-build-${{ env.cache-name }}-
               ${{ runner.os }}-build-
               ${{ runner.os }}-
-        - name: Build package in rattler-build (experimental)
+        - name: Convert recipes and dry-run rattler-build
           run: |
             source $CONDA/bin/activate
             conda activate conda-recipe-manager
             conda install -y -c conda-forge rattler-build
-            convert tests/test_aux_files/integration/${{ matrix.test-directory }}
+            convert -m 0.80 -o recipe.yaml tests/test_aux_files/integration/${{ matrix.test-directory }}
+            rattler-bulk-build -m 0.30 tests/test_aux_files/integration/${{ matrix.test-directory }} --render-only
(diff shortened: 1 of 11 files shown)


.github/workflows/ci.yaml
```diff
@@ -64,22 +52,30 @@ jobs:
         - uses: ./.github/actions/setup-env
           with:
             python-version: "3.11"
-        - name: Cache conda env
```
schuylermartin45, line 67: Ignoring the caching issues I ran into with the integration build, I briefly looked at the past runs and found that this caching phase adds more time. So I've removed all CI conda caching in this project.

conda_recipe_manager/commands/rattler_bulk_build.py
```diff
@@ -0,0 +1,162 @@
...
+            "thread_pool_size": thread_pool_size,
+        },
+    }
+    final_output = {
+        "recipes_with_build_error_code": recipes_with_errors,
+        "error_histogram": error_histogram,
+        "stats": stats,
+    }
+
+    print(json.dumps(final_output, indent=2))
```
jezdez, line 161: Do you want to print something to stderr in case we miss the success threshold?

jezdez (approved):

conda_recipe_manager/commands/rattler_bulk_build.py
schuylermartin45: I've thought about that but I think I want to reserve STDERR for non-JSON logging. Since the CI pipeline will dump the JSON log with the percentage AND send a failed exit code, I feel like writing another message is a bit redundant.

PR #15, opened 2024-03-27
Adds bioconda recipe conversion integration test

schuylermartin45:
- All bioconda recipes (as of 2024-03-27) has been copied into this project for the purposes of integration testing the conversion work with rattler-build
- This will be a great way to test both the recipe conversion AND determine if the other integration tests can scale to a few thousand recipes
- Hopefully the bioconda folks won't mind me doing this...

UPDATE: Half of the bioconda recipes are currently not being evaluated. I have raised #16 to figure out why. For now, the files are included but the tests are disabled.


baszalmstra:
Instead of adding all these files to this repository (making it huge), maybe you could dynamically fetch them or use a git submodule?

schuylermartin45:
I want the integration tests to be snapshots and stable. In principle, the tests should only fail IFF there is an issue with the conversion work or rattler-build. The success metrics per test are expected to increase over time as compatibility improves.

Fetching individual files would mean a ton of HTTP requests are made per run at this scale AND I'd have to mark each commit hash I'd want to fetch to ensure the files do not change between tests. That doesn't sound very manageable at this scale.

Submodules have the same fetch issue don't they? Not to mention I'd have to fetch the whole feedstock repo at once and not just the one file I am interested in? In my testing a few thousand recipes only take a few dozen megabytes of disk space, whereas a few thousand feedstock repositories take up potentially gigabytes of disk, and it's potentially unbounded.

schuylermartin45:
I've also considered putting the integration test files in a separate repo to checkout in the workflow, but that effectively has the same network fetch issue, albeit a bit easier to manage.

schuylermartin45:
On top of all this, I generally try to avoid the network in my automated testing. It adds a layer of uncertainty and another point of failure outside of the domain of the work being tested.

baszalmstra:
I understand where you are coming from. 👍 

I do feel like this adds a lot of bloat to the repository. I would personally use a submodule which in my eyes is at the same level harmful as the git repository itself in terms of netwerk access. But I do understand that you want to have more control over the files you include.

But disregard my comment, having these integration tests is awesome. :)

schuylermartin45:
No I'm glad you brought it up, I have had some of these thoughts/conversations with others internally.

It sucks to have thousands of files in a repo but I don't think I've seen/can think of an alternative that manages the pros/cons better than this.

jezdez:
@schuylermartin45 This is a no-go, let's use a git submodule for this instead of copying over those files.

schuylermartin45:
Let's talk on early next week. I have some other ideas that I think are best discussed over the phone.

schuylermartin45:
I'm going to be closing this PR and it's sister draft PR. We'll be moving the integration test files into it's own separate repo to checkout/cache in the CI workflow. So that'll be 1 potential network call over having to manage thousands, with no need to mess with submodules.

In the mean time, I will be cleaning up the branch from the draft PR and re-submit it as a new PR with the new scripts, but no test files.

PR #22, opened 2024-04-08
Refactors recipe conversion work

schuylermartin45:
- Recipe conversion tooling is now broken-out into a subclass of the original `RecipeParser`
- This keeps the `RecipeParser file from getting any bigger and will allow some future flexibility in maintaining the conversion process
- Test files are broken up accordingly

**There are no net logic changes, only code has been moved.**


tests/parser/test_recipe_parser.py
```diff
@@ -1,22 +1,18 @@
...
 
 from conda_recipe_manager.parser.enums import SelectorConflictMode
 from conda_recipe_manager.parser.exceptions import JsonPatchValidationException
 from conda_recipe_manager.parser.recipe_parser import RecipeParser
-from conda_recipe_manager.parser.types import MessageCategory
 from conda_recipe_manager.types import JsonType
-
-# Path to supplementary files used in test cases
-TEST_FILES_PATH: Final[str] = "tests/test_aux_files"
+from tests.conftest import TEST_FILES_PATH, load_file, load_recipe
```
jezdez, line 15: Pytest conftest modules shouldn't be imported directly, it's better to move that into a tests/file_loading.py or something that you can import, or define it as a fixture that takes request parameters via the ["Indirect parametrization" feature](https://docs.pytest.org/en/7.1.x/example/parametrize.html#indirect-parametrization).

jezdez (changes requested):
Just a small change in how you define these helpers, they would work well as fixtures with indirect parameters, or as utility functions living outside of the magic conftest module.

tests/parser/test_recipe_parser.py
schuylermartin45: I'll just rename the file for now, to keep it simple. I think we have other projects in Distro that mis-use `conftest.py` in that case.

jezdez:
@schuylermartin45 I think the file is missing 🤗

schuylermartin45:
> @schuylermartin45 I think the file is missing 🤗

This is what I get for quickly doing `git add -u` from mussel memory. Let me fix that.

jezdez (commented):
The tests seem to have trouble importing the new module, not sure why it wouldn't, maybe the indirection via make?

schuylermartin45:
I'm not sure if I follow, it looks like the only test failing is the rattler-build test that's probably a rattler issue.

marcelotrevisani:
> I'm not sure if I follow, it looks like the only test failing is the rattler-build test that's probably a rattler issue.

it is not really a problem with rattler, that is due the name path being too long
https://docs.python.org/3/library/errno.html#errno.ENAMETOOLONG

if you can try simply reduce the name of the file, that should do it

tests/file_loading.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    Convenience function that simplifies initializing a recipe parser.
+    :param file_name: File name of the test recipe to load
+    :returns: RecipeParser instance, based on the file
+    """
+    recipe = load_file(f"{TEST_FILES_PATH}/{file_name}")
+    return RecipeParser(recipe)
+
+
+def load_recipe_convert(file_name: str) -> RecipeParserConvert:
```
marcelotrevisani, line 37: to load files for tests you can use ``pytest-datadir`` if you wish, that would simplify a few things

tests/file_loading.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    Loads a file into a single string
+    :param file: Filename of the file to read
+    :returns: Text from the file
+    """
+    with open(Path(file), "r", encoding="utf-8") as f:
+        return f.read()
+
+
+def load_recipe(file_name: str) -> RecipeParser:
```
marcelotrevisani, line 27: if you prefer you can use factories as fixtures here and remove the ``__init__.py`` and this bit here, and put everything in the ``conftest.py``
example:
https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#factories-as-fixtures

tests/file_loading.py
```diff
@@ -0,0 +1,44 @@
...
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Final
+
+from conda_recipe_manager.parser.recipe_parser import RecipeParser
+from conda_recipe_manager.parser.recipe_parser_convert import RecipeParserConvert
+
+# Path to supplementary files used in test cases
+TEST_FILES_PATH: Final[str] = "tests/test_aux_files"
```
marcelotrevisani, line 14: this bit here will work only if you execute the tests from the root repository.
I would recommend getting it relative of the test file maybe? or using ``pytest-datadir``
something like

```python
TEST_FILES_PATH = Path(__file__).parent / "test_aux_files"
```

schuylermartin45:
> > I'm not sure if I follow, it looks like the only test failing is the rattler-build test that's probably a rattler issue.
> 
> it is not really a problem with rattler, that is due the name path being too long https://docs.python.org/3/library/errno.html#errno.ENAMETOOLONG
> 
> if you can try simply reduce the name of the file, that should do it

I've been pinning it on rattler as I think it is a change between version. That path is not new and has worked in previous CI builds.
But again, I haven't had time today to investigate this further.

tests/file_loading.py
schuylermartin45: I appreciate and am aware, but I wanted to keep this PR smaller and focus refactoring one thing.

tests/file_loading.py
schuylermartin45: TIL, I'll have to look into that. I was not aware of that plugin.

tests/file_loading.py
schuylermartin45: Fair point. I almost always run the tests from the `make` directives so I haven't personally had that issue.

schuylermartin45:
> I've been pinning it on rattler as I think it is a change between versions. That path is not new and has worked in previous CI builds. But again, I haven't had time today to investigate this further.

Well good news, I finally found time to look into :). From what I can tell, my initial gut feeling looks correct and [I've filed an issue with rattler-build](https://github.com/prefix-dev/rattler-build/issues/778). Version `0.13.0` did not have an issue with this file path.

tests/file_loading.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def load_file(file: Path | str) -> str:
+    """
+    Loads a file into a single string
+    :param file: Filename of the file to read
+    :returns: Text from the file
+    """
+    with open(Path(file), "r", encoding="utf-8") as f:
+        return f.read()
```
kenodegard, line 24: `pathlib` offers an even simpler way of reading a text file:

```suggestion
    Path(file).read_text()
```

tests/file_loading.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def load_recipe(file_name: str) -> RecipeParser:
+    """
+    Convenience function that simplifies initializing a recipe parser.
+    :param file_name: File name of the test recipe to load
+    :returns: RecipeParser instance, based on the file
+    """
+    recipe = load_file(f"{TEST_FILES_PATH}/{file_name}")
```
kenodegard, line 33: `TEST_FILES_PATH` is already a `Path` object, let's use it as such:

```suggestion
    recipe = load_file(TEST_FILES_PATH / file_name)
```

tests/file_loading.py
```diff
@@ -0,0 +1,44 @@
...
+    return RecipeParser(recipe)
+
+
+def load_recipe_convert(file_name: str) -> RecipeParserConvert:
+    """
+    Convenience function that simplifies initializing a recipe parser.
+    :param file_name: File name of the test recipe to load
+    :returns: RecipeParserConvert instance, based on the file
+    """
+    recipe = load_file(f"{TEST_FILES_PATH}/{file_name}")
```
kenodegard, line 43:
```suggestion
    recipe = load_file(TEST_FILES_PATH / file_name)
```

conda_recipe_manager/parser/recipe_parser_convert.py
```diff
@@ -0,0 +1,360 @@
...
+from conda_recipe_manager.parser._utils import stack_path_to_str, str_to_stack_path
+from conda_recipe_manager.parser.recipe_parser import RecipeParser
+from conda_recipe_manager.parser.types import CURRENT_RECIPE_SCHEMA_FORMAT, MessageCategory, MessageTable
+from conda_recipe_manager.types import JsonPatchType, JsonType, SentinelType
+
+
+class RecipeParserConvert(RecipeParser):
+    """
+    Extension of the base RecipeParser class to enables upgrading recipes from the old to new format.
+    This was originally part of the RecipeParser class but was broken-out for easier maintenance.
```
kenodegard, line 28: How does breaking this out like this offer easier maintenance?

As far as I can tell the maintenance issue here is caused by the high cyclomatic complexity of the `render_to_new_recipe_format` method. A good rule of thumb is to aim for a complexity of 15 or lower but, as currently written, this method has a complexity of 49, much too high.

Furthermore if it is this easy to break this into a new class (and since `self` is only referenced once in the method) it would seem to me that this deserves to live as a function and not nested within a larger object:

```python
def convert_to_new_recipe_format(old_recipe: RecipeParser) -> tuple[str, MessageTable]:
    new_recipe = RecipeParser(self.render())
    ...
```

tests/file_loading.py
schuylermartin45: Good catch! Missed that change when I switched it from a string.

conda_recipe_manager/parser/recipe_parser_convert.py
schuylermartin45: The subclass is mostly there to make the static analyzer happy with accessing underscore functions. I also have future plans to break this up into more functions under this new "namespace".

As for maintenance, the original parser class is huge. It used to be an even larger class file and was split-up into a few supporting files and classes.

Out of curiosity, how did you calculate that? Do you have a plugin? I also think some of this logic is unavoidable with the nature of the conversion work. There is a lot of "if this, then that" to perform.

schuylermartin45: Let me go and clean up a bit more here. I think I can have the message tracking and a few other pieces of state tracked as part of the class member that will simplify _some_ of the ick.

kenodegard: I have McCabe linters enabled in my IDE. Since you're using `pylint` you can use https://pypi.org/project/pylint-mccabe/.

schuylermartin45: I just posted a few commits that clean-up the inner-function and state usage. If you're okay with it, I'd like to look into this `mccabe` integration after this PR and #29 get merged in. I think that might cause a bigger refactor

conda_recipe_manager/commands/convert.py
```diff
@@ -2,6 +2,7 @@
 File:           convert.py
 Description:    CLI for converting an old recipe file to the "new" format.
 """
+
```
schuylermartin45, line 5: `black` got an update in `pre-commit` the other day, hence the change here.

kenodegard (approved):

PR #17, opened 2024-04-01
conda-forge Integration Tests

schuylermartin45:
This is a DRAFT PR, predicated on acceptance of #15 

This contains 2000 randomly selected conda-forge recipe files to use for integration testing. Also includes the script used to procure the test data.

Putting this PR up now to get some preliminary test results.

PR #26, opened 2024-04-08
Integration Testing, Round 2

schuylermartin45:
This PR supersedes #15 and #17 and contains the work of #17, with the additions of thousands of test files removed. 

This PR will remain a draft until we can bring the new repo of test data online.


schuylermartin45:
This is now ready for review, but we should get #22 and #29 merged in first (for bug fixes and other improvements).
This currently does not use a cache, but it is using the `sparse` checkout option. Each instance only pulls the testing files required for each individual test. So far that seems to be fast enough, with tests only taking 2 minutes to succeed.

PR #29, opened 2024-04-10
Refactor: Commands execute under `conda-recipe-manager` namespace

schuylermartin45:
- Addresses feedback from https://github.com/conda-incubator/conda-recipe-manager/issues/19
  - `convert` and `rattler-bulk-build` now execute under `conda-recipe-manager`
      - Alternatively, use `crm` for short
  - For example, instead of `convert recipe/meta.yaml` you now use:
    `conda-recipe-manager convert recipe/meta.yaml`
- Refactors and adds missing CLI smoke tests
- Breaks-out duplicated work in the scripts into a new `commands.utils.print`
  module
- `black` got an update that changes spacing between `from __future__ ...` and
  the file comment that precedes it

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,4 +1,4 @@
 # Syntax for this file at https://help.github.com/articles/about-codeowners/
 
-*       @conda/builds-tools
+*       @conda-incubator/builds-tools
 *       @schuylermartin45
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,4 +1,3 @@
 # Syntax for this file at https://help.github.com/articles/about-codeowners/
 
-*       @conda-incubator/builds-tools
-*       @schuylermartin45
+*   @conda-incubator/builds-tools @schuylermartin45
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,4 @@
+# Syntax for this file at https://help.github.com/articles/about-codeowners/
+
+*       @conda/builds-tools
+*       @schuylermartin45
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,3 +1,3 @@
 # Syntax for this file at https://help.github.com/articles/about-codeowners/
 
-*   @conda-incubator/builds-tools @schuylermartin45
+*   @conda-incubator/builds-tools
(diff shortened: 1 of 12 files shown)


schuylermartin45:
`build-recipe-rattler` issue is resolved in #22

schuylermartin45:
Testing if a close/open will trigger `CODEOWNERS` changes.

.github/CODEOWNERS
kenodegard, line 1: I suggest we cherrypick this change into a separate PR.

Also since you are a member of the builds-tools team you do not need to explicitly include yourself.

tests/parser/test_recipe_parser.py
```diff
@@ -19,9 +20,11 @@
 TEST_FILES_PATH: Final[str] = "tests/test_aux_files"
 
 # Long multi-line description string found in the `simple-recipe.yaml` test file
-SIMPLE_DESCRIPTION: Final[
-    str
-] = "This is a PEP '561 type stub package for the toml package.\nIt can be used by type-checking tools like mypy, pyright,\npytype, PyCharm, etc. to check code that uses toml."  # pylint: disable=line-too-long
+SIMPLE_DESCRIPTION: Final[str] = (
+    "This is a PEP '561 type stub package for the toml package."
+    "\nIt can be used by type-checking tools like mypy, pyright,"
+    "\npytype, PyCharm, etc. to check code that uses toml."
```
kenodegard, line 26: FYI in other projects we've preferred to leave the newlines at the end

```suggestion
    "This is a PEP '561 type stub package for the toml package.\n"
    "It can be used by type-checking tools like mypy, pyright,\n"
    "pytype, PyCharm, etc. to check code that uses toml."
```

kenodegard (changes requested):
Almost ready to go, two minor suggestions

.github/CODEOWNERS
schuylermartin45: I'll redact my name. I had a reason for it but now I don't remember, probably just paranoia.

schuylermartin45: As for the separate PR, why? It's a small change and I'd like to have it in sooner than later.

kenodegard: a small separate PR with such a highly focused changed like that would've already been approved and merged by now

kenodegard (approved):

.github/CODEOWNERS
schuylermartin45: Lol I suppose so, maybe we can chat offline sometime. I'm sure this just comes from a difference in team experiences.

Issue #25, opened 2024-04-08
Selectors in a list are not being converted properly


schuylermartin45:
From the conda-build element chat:

```yaml
  build:
    - python                                 # [build_platform != target_platform]
    - cross-python_{{ target_platform }}     # [build_platform != target_platform]
```
is not quite corretly translated (only the first of the two elements makes it into the `then` part of the condition.

schuylermartin45:
Confirmed, here are the results with a very basic recipe file:
Input:
```yaml
build:
  - python                                 # [build_platform != target_platform]
  - cross-python_{{ target_platform }}     # [build_platform != target_platform]
```

Output:
```yaml
schema_version: 1

context:

build:
  - if: build_platform != target_platform
    then: python
  - cross-python_${{ target_platform }}
```

PR #33, opened 2024-04-17
Fixes Issue #25

schuylermartin45:
- Fixes an issue described in #25 that left some selectors unconditionalized.
- To address this, selectors are converted first before any other operations are performed. By doing this phase first, we ensure that the selector table points to the correct paths and no selectors are accidentally wiped by patch operations.
- Makes some minor refactoring for the sake of code clean-up.
- This does not add any new regression tests, BUT does fix existing tests to ensure that the regression is caught. The mistake was previously missed when adding those tests.


schuylermartin45:
@wolfv  for visibility.

beeankha (approved):

PR #37, opened 2024-04-22
Logs are now automatically aggregated into the GHA Summary

schuylermartin45:
Automates the aggregation of integration test statistics from the various integration tests that run as part of this.
This process only runs if all integration tests pass (which are required to pass to merge a PR).

Also:
- Splits-up the `ci.yaml` file into two files, one for "commit checks" and one for integration tests
- Fixes some warning messages
- Bumps the latest integration test thresholds.

To view the summary of results, go to this repo on GitHub, `Actions` -> `Workflow Run` (instance in the list) -> `Summary`.


.github/actions/setup-env/action.yaml
```diff
@@ -1,6 +1,12 @@
...
 
+inputs:
+  python-version:
+    description: "Version of python to use in the conda environment."
+    required: false
+    default: "3.11"
+
 runs:
   using: "composite"
   steps:
```
jezdez, line 12: This shouldn't use setup-python, but instead of setup-miniconda to make sure we have a consistent Python build in use. In fact, you'll likely be able to get rid of this setup-env workflow completely since setup-miniconda implements the steps except for the `pip install -e .` call.

jezdez (changes requested):
Let's use this oppurtuntiy to switch to setup-miniconda from a mixed setup-python (which installs python.org's Python) and relying on GitHub Actions' conda.

.github/actions/setup-env/action.yaml
schuylermartin45: I setup-python for setup-miniconda. I was hopping to get some more of the boiler-plate out of the way, but my experiments did not pan out. The point of this action is to reduce duplicated start-up work across all the workflows, so I'm okay keeping it. It makes it easier if I have to change the setup steps in the future.

beeankha (approved):
Walked through this CI infrastructure work with @schuylermartin45 , LGTM!

Issue #36, opened 2024-04-19
Inclusive language conversions


schuylermartin45:
After a brief discussion with @beeankha , we noticed that the new recipe format replaces outdated terminology for new inclusive language in a handful of places, primarily with allow/deny lists in recipe files.

I don't think the current conversion work addresses these changes yet, but we will need to in order to pass the schema checks in `rattler-build`.

PR #38, opened 2024-04-25
Inclusive language fixes, bug fixes, a small refactor, and more unit tests

schuylermartin45:
Addresses the concerns raised in #36 (which also moves us towards better compliance with CEPs 13 & 14.

Also
- Fixes a bug that this work uncovered that relates to YAML parsing dropping quotes around strings that need quotes. Previously we had a partial fix for this edge case. This new fix attempts to address the issue at parse-time.
- Conversion tooling now has another convenience operation to streamline a common move operation.
- Adds more unit tests, one of which is a WIP


beeankha (approved):

wolfv:
Awesome!

PR #54, opened 2024-05-07
Quick bug fixes

schuylermartin45:
A few quick bug fixes I found and fixed today.

Also includes some minor spelling corrections that happen on occasion. They don't happen a lot but it's weird they happen more than once.

Also fixes an issue where the integration tests may silently fail to meet the testing threshold. The `pipefail` shell option was not enabled, causing `tee` to always return `0`

### DIFF
--- a/conda_recipe_manager/parser/recipe_parser_convert.py
+++ b/conda_recipe_manager/parser/recipe_parser_convert.py
@@ -238,7 +238,7 @@ class RecipeParserConvert(RecipeParser):
 
     def _upgrade_build_section(self, base_package_paths: list[str]) -> None:
         """
-        Upgrades/converts the `about` section(s) of a recipe file.
+        Upgrades/converts the `build` section(s) of a recipe file.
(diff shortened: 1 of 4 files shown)


conda_recipe_manager/parser/recipe_parser_convert.py
```diff
@@ -197,6 +197,17 @@ def _upgrade_selectors_to_conditionals(self) -> None:
                 self._patch_and_log(patch)
                 self._v1_recipe.remove_selector(selector_path)
 
+    def _correct_common_misspellings(self) -> None:
+        """
+        Corrects common spelling mistakes in field names.
+        """
+        # "If I had a nickel for every time `skip` was misspelled, I would have several nickels. Which isn't a lot, but
+        #  it is weird that it has happened multiple times."
+        #                                                             - Dr. Doofenshmirtz, probably
```
beeankha, line 206: Ha! 😆

beeankha (approved):
👍

PR #55, opened 2024-05-08
JINJA substitution fixes

schuylermartin45:
I found some scenarios where the JINJA substitution upgrades can fail and cause doubling up of the dollar sign (i.e. `$${{ foo }}` is added to the file instead of `${{ foo }}`.

This PR addresses those issues and adds regression tests to prevent this from happening again.

Also:
- Adds some new debug logging options to `rattler-bulk-build`
- Adds some simple license corrections to the SPDX format

### DIFF
--- a/conda_recipe_manager/parser/recipe_parser_convert.py
+++ b/conda_recipe_manager/parser/recipe_parser_convert.py
@@ -150,7 +150,8 @@ class RecipeParserConvert(RecipeParser):
                     MessageCategory.WARNING, f"A non-string value was found as a JINJA substitution: {value}"
                 )
                 continue
-            value = value.replace("{{", "${{")
+            # Safely replace `{{` but not any existing `${{` instances
+            value = Regex.JINJA_REPLACE_V0_STARTING_MARKER.sub("${{", value)
             self._patch_and_log({"op": "replace", "path": path, "value": value})
 
     def _upgrade_selectors_to_conditionals(self) -> None:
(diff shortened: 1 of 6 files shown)


conda_recipe_manager/parser/recipe_parser_convert.py
```diff
@@ -309,6 +313,30 @@ def _upgrade_requirements_section(self, base_package_paths: list[str]) -> None:
...
+
+        if old_license_sanitized not in SPDX_COMMON_MISSPELLINGS_TBL:
+            return
+
+        new_license: Final[str] = SPDX_COMMON_MISSPELLINGS_TBL[old_license_sanitized]
+        # Alert the user that a patch was made, in case it needs manual verification
+        if self._patch_and_log({"op": "replace", "path": license_path, "value": new_license}):
+            self._msg_tbl.add_message(
+                MessageCategory.WARNING, f"Changed {license_path} from `{old_license}` to `{new_license}`"
+            )
```
jezdez, line 338: I'd use something like [`difflib.get_close_matches`](https://docs.python.org/3/library/difflib.html#difflib.get_close_matches) or similar to make this a little smarter

jezdez (approved):

conda_recipe_manager/parser/recipe_parser_convert.py
schuylermartin45: I had it in my head that I'd have to make a network request to do any sort of general comparison. This small look-up table fixes about 30% of our current issues.

But now I've just found that SPDX publishes a handful of data tables on GitHub, including a JSON file:
https://github.com/spdx/license-list-data/blob/main/json/licenses.json

We'll need a patch table for some of the mismatches, but I'll take a look into this route for the next PR.

PR #65, opened 2024-05-24
Adds `script_env` support

### LINKED ISSUE #64: `/build/script` needs to be upgraded
The `/build/script` section is currently not being upgraded. In simple scenarios, this works out fine. In complex situations, we need to use the new [Script object](https://github.com/conda/ceps/blob/main/cep-14.md#script-section)

This will be a pretty involved upgrade process that will need to be heavily conditionalized. From a brief search, there are already many variants AND we have to parse-out shell variable syntax in the existing [script_env](https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#use-environment-variables) field.

schuylermartin45:
- Adds basic support for upgrading a recipe that contains a `script_env` section.
  - NOTE: There is one edge case that is not currently supported and is likely best left for a human to fix. The conversion process will flag this with an error.
- Adds unit tests

Fixes #64

### DIFF
--- a/conda_recipe_manager/parser/_utils.py
+++ b/conda_recipe_manager/parser/_utils.py
@@ -16,7 +16,7 @@ from conda_recipe_manager.parser._types import (
     StrStackImmutable,
 )
 from conda_recipe_manager.parser.types import TAB_AS_SPACES, MultilineVariant, NodeValue
-from conda_recipe_manager.types import H, SentinelType
+from conda_recipe_manager.types import H, JsonType, SentinelType
 
 # Commonly used special characters that we need to ensure get quoted when rendered as a YAML string.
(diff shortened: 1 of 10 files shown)


conda_recipe_manager/parser/_utils.py
```diff
@@ -188,3 +188,15 @@ def dedupe_and_preserve_order(l: list[H]) -> list[H]:
 
     """
     return list(cast(dict[H, None], dict.fromkeys(l)))
+
+
+def set_key_conditionally(dictionary: dict[str, JsonType], key: str, value: JsonType) -> None:
+    """
+    Convenience function that conditionally includes a key-value pair in a dictionary if the value is truthy.
+    Great for cheating McCabe ratings in complex JSON/YAML operations!
```
beeankha, line 196: 😆

beeankha (approved):

PR #72, opened 2024-06-11
Upgrades basic quoted multiline strings

schuylermartin45:
A handful of recipes have quoted multiline strings, that the parser does not support:
```yaml
  description: "Tests handling of\nmultiline quoted strings"
```

So the preprocessing phase will now attempt to convert this to a pipe-marked multiline string:
```yaml
  description: |
    Tests handling of
    multiline quoted strings
```

There aren't many recipes that use this format, and even fewer that have multiple newlines in the quoted string. Hence why this is a pre-processor fix.

### DIFF
--- a/conda_recipe_manager/parser/_types.py
+++ b/conda_recipe_manager/parser/_types.py
@@ -136,7 +136,11 @@ class Regex:
         r"(\{\{\s*[a-zA-Z0-9_]*.*)\.([a-zA-Z0-9_]*\(.*\)\s*\}\})"
     )
     # Strips empty parenthesis artifacts on functions like `| lower`
-    PRE_PROCESS_JINJA_DOT_FUNCTION_STRIP_EMPTY_PARENTHESIS = re.compile(r"(\|\s*(lower|upper))(\(\))")
+    PRE_PROCESS_JINJA_DOT_FUNCTION_STRIP_EMPTY_PARENTHESIS: Final[re.Pattern[str]] = re.compile(
+        r"(\|\s*(lower|upper))(\(\))"
+    )
+    # Attempts to normalize multiline strings containing quoted escaped newlines.
+    PRE_PROCESS_QUOTED_MULTILINE_STRINGS: Final[re.Pattern[str]] = re.compile(r"(\s*)(.*):\s*['\"](.*)\\n(.*)['\"]")
 
     ## Jinja regular expressions ##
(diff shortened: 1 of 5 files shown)


conda_recipe_manager/parser/_types.py
```diff
@@ -136,7 +136,11 @@ class Regex:
         r"(\{\{\s*[a-zA-Z0-9_]*.*)\.([a-zA-Z0-9_]*\(.*\)\s*\}\})"
     )
     # Strips empty parenthesis artifacts on functions like `| lower`
-    PRE_PROCESS_JINJA_DOT_FUNCTION_STRIP_EMPTY_PARENTHESIS = re.compile(r"(\|\s*(lower|upper))(\(\))")
+    PRE_PROCESS_JINJA_DOT_FUNCTION_STRIP_EMPTY_PARENTHESIS: Final[re.Pattern[str]] = re.compile(
```
jezdez, line 139: That reminds me btw that I wonder if it would make a performance dent to use [Google's re2](https://github.com/conda-forge/google-re2-feedstock)?

jezdez (approved):

conda_recipe_manager/parser/_types.py
schuylermartin45: https://github.com/google/re2/wiki/WhyRE2
> It is not a goal to be faster than all other engines under all circumstances. Although RE2 guarantees linear-time performance, the linear-time constant varies depending on the overhead entailed by safe handling of the regular expression. In a sense, RE2 behaves pessimistically whereas backtracking engines behave optimistically, so it can be outperformed in various situations.
Sounds like the authors acknowledge it may not be faster in all scenarios.

> It is also not a goal to implement all of the features offered by Perl, PCRE and other engines. As a matter of principle, RE2 does not support constructs for which only backtracking solutions are known to exist. Thus, backreferences and look-around assertions are not supported.
I believe we have a number of backreference expressions, so this would not be a drop-in replacement.

I do not have performance tests on this project. Single recipes are converted very quickly, sub-second most of the time in my experience. We can convert 2,000 on the base GHA instance in ~3 minutes (including instance spin-up/shut-down time).

Our time would be better spent improving/further automating manual human tasks. If we wanted conversion to go faster, we'd rewrite this project in Rust first.

jezdez: Good points, thanks for taking a look!

PR #77, opened 2024-07-08
Adds Build Integration Test

schuylermartin45:
Adds some full-build integration tests

PR #80, opened 2024-07-16
Upgrades commonly used selectors

schuylermartin45:
This discovery came out of developing #77 and some discussions held on the public Conda Build Tools Element channel.

The commonly used `py<37` selector expression is no longer supported in the new recipe format (see https://github.com/conda/ceps/pull/71 and https://prefix-dev.github.io/rattler-build/latest/selectors/#selector-evaluation).

These selector aliases/expressions have been superseded by the new `match()` function in the V1 recipe format.

This is the first issue we have found that occurs in when performing a full build with `rattler-build` BUT does not appear in a dry-run build. There are likely many more such issues to be discovered.

This PR also stores `/build/skip` expressions as a single value instead of a list. I believe this was a temporary workaround to a schema issue that has since been resolved in `rattler-build`.


wolfv:
Looks great!

wolfv (approved):

PR #81, opened 2024-07-17
Adds some basic support for editing V1 recipe files

schuylermartin45:
It has occurred to me that it shouldn't take a lot of work to make the recipe parser capable of editing both V0 and V1 recipe files. You can already make some modifications to V1 recipe files, but this is untested and there will be bugs.

This is the first PR of many that will allow us to edit both V0 and V1 recipes programmatically. The parser is now able to detect which format of a recipe it has been given. Using that knowledge, it will branch on logic that is version specific.

To start, this PR fixes logic around JINJA variable management and starts adding unit test coverage for V1 recipes.

This unit test coverage is not extensive, but I wanted to break this work into multiple PRs before it got too big.

### DIFF
--- a/conda_recipe_manager/commands/convert.py
+++ b/conda_recipe_manager/commands/convert.py
@@ -18,6 +18,7 @@ from typing import Final, Optional
 import click
 
 from conda_recipe_manager.commands.utils.print import print_err, print_messages, print_out
+from conda_recipe_manager.parser.enums import SchemaVersion
 from conda_recipe_manager.parser.recipe_parser_convert import RecipeParserConvert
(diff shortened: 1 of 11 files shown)


dholth:
I guess nobody including conda-forge parses v1 recipes into a real syntax tree including comments and jinja stuff?

schuylermartin45:
> I guess nobody including conda-forge parses v1 recipes into a real syntax tree including comments and jinja stuff?

I'm not currently aware of any project capable of editing V1 recipes programmatically like this. I think it is worth the time to support given how close `conda-recipe-manager` is already at parsing V1 files.

dholth:
Super cool

PR #90, opened 2024-08-01
Platform Selector Parsing for V0 Recipe Files

schuylermartin45:
This PR adds a new class that is capable of parsing selector statements, without evaluating Python statements as code.

For now the purpose of this work is to determine which Platforms (OS + Architecture combinations) apply to a given selector. Future work will use this to determine which recipes are `noarch` candidates.

In the future, we may also expand on these capabilities to determine which Python versions a recipe may be apply to.

Also introduces a new `IsModifiable` base class to reduce repeated code in parsers.

Note that not all Platforms/Architectures/Operating Systems recognized by conda are covered at this time.

@markan & @jmlphd for visibility


schuylermartin45:
It looks like I accidentally used some 3.12 features not available in 3.11 which are breaking the tests.

As a side note, I think we're at the point where we can add 3.12 to the test matrix.

conda_recipe_manager/parser/platform_types.py
```diff
@@ -0,0 +1,162 @@
...
+
+from __future__ import annotations
+
+from enum import StrEnum
+from typing import Final
+
+
+class OperatingSystem(StrEnum):
+    """
+    Operating System enumeration. This a broad (and sometimes inaccurate) qualifier supported by the recipe format.
```
beeankha, line 17:
```suggestion
    Operating System enumeration. This is a broad (and sometimes inaccurate) qualifier supported by the recipe format.
```

tests/parser/test_selector_parser.py
```diff
@@ -0,0 +1,129 @@
...
+                # OSX is included in the UNIX category, therefore it is included
+                Platform.OSX_64,
+                Platform.OSX_ARM_64,
+                Platform.WIN_32,
+                Platform.WIN_64,
+                Platform.WIN_ARM_64,
+            },
+        ),
+        ("[osx and not unix]", SchemaVersion.V0, set()),
+        # TODO FIX: Python versions should have no affect on which Platforms are included
```
beeankha, line 102:
```suggestion
        # TODO FIX: Python versions should have no effect on which Platforms are included
```

beeankha (approved):
Just some small nitpicky comment/docstring edits, otherwise looks good to me!

PR #93, opened 2024-08-15
Advanced Dependency RecipeParser Tooling

schuylermartin45:
- Introduces a two new dependency functions to the recipe parser module.
  - To keep the Recipe Parser from growing even more, these functions are added in a new sub-class called `RecipeParserDeps`
- These tools help identify, parse, and organize all dependencies found in a single or multi-output recipe file.
- These tools should allow us to better statically analyze dependencies in a recipe file.
- Adds `==` operator to `SelectorParser` class
- Adds many unit tests for the work described above ^

For visibility:
@JeanChristopheMorinPerso @markan


conda_recipe_manager/parser/dependency.py
```diff
@@ -0,0 +1,108 @@
...
+    # Identifies what kind of dependency this is
+    type: DependencySection
+    # Parses the dependency's name and version constraints
+    match_spec: MatchSpec
+    # The selector applied to this dependency, if applicable
+    selector: Optional[SelectorParser] = None
+
+
+# Maps-out dependencies found in a recipe. Maps package name -> list of parsed dependencies.
+DependencyMap = dict[str, list[Dependency]]
```
schuylermartin45, line 108: This structure contains a list of dependencies for the sake of simplicity. I figured for most of our use cases, we will end up iterating over all the dependencies anyways. At the very least, we'll likely need to filter one or more `DependencySection`s at a time from this list.

So I thought a `dict[str, dict[DependencySection, list[Dependency]]` would just end up causing more headaches and would not buy us much in terms of performance. Also I haven't seen a recipe file with thousands of listed dependencies so the search space is small.

conda_recipe_manager/parser/dependency.py
```diff
@@ -0,0 +1,109 @@
...
+class DependencySection(Enum):
+    """
+    Enumerates dependency sections found in a recipe file.
+    """
+
+    BUILD = auto()
+    HOST = auto()
+    RUN = auto()
+    RUN_CONSTRAINTS = auto()
+    # NOTE: This was not in the `requirements/` section in V0
```
beeankha, line 25: Sorry to nitpick, but I think the placement of this comment doesn't indicate which requirements section wasn't in V0 (to a new person, they might be like "Is it the section above, or the section below?"); I would suggest either putting the comment on the same line or specify what the new section is directly in the comment.

schuylermartin45: Usually a comment refers to the line directly below it. That being said, I think it would read better replacing `This`. Thanks for the call-out!

beeankha (approved):

PR #106, opened 2024-08-27
add patch command 

### LINKED ISSUE #83: `crm patch <JSON patch> <file>` Command
To enable experimentation and shell-scripting small recipe edits, I would like to propose a simple CLI command to edit recipe files.

Given a file and a JSON-patch blob, this command will apply an edit to a recipe file. This command should validate the patch blob and recipe file provided.

Examples:
```sh
$ crm patch {"op": "not_an_op", "foo": "/about/license", "value": "MIT"} /foo/bar/meta.yaml
ERROR: Illegal JSON patch provided

$ crm patch {"op": "add", "path": "/about/license", "value": "MIT"} /foo/bar/not_a_recipe.txt
ERROR: Failed to recognize `/foo/bar/not_a_recipe.txt` as a recipe file.

$ crm patch {"op": "add", "path": "/about/license", "value": "MIT"} /foo/bar/meta.yaml
Success! Changes have been saved to `/foo/bar/meta.yaml`
```

Bonus points: provide a diff of the changes made.
```sh
$ crm patch --diff {"op": "add", "path": "/about/license", "value": "MIT"} /foo/bar/meta.yaml
Success! Changes have been saved to `/foo/bar/meta.yaml`
about:
  summary: foo
------42-42
  license: Apache 2.0
+++++42-42
  license: MIT
```
There are likely a few other flags we can add to make this more useful for users. Maybe log the file to STDOUT instead of writing to it directly?

**Mahe's Notes**
instead of having json strings in terminal, it is okay to have the path to a json file in the first iteration.
Break down of what needs to be done:

- [ ] Add patch command to help message
- [ ] Study the "make format", "make pre-commit" commands
- [ ] validate if the file is a valid recipe or not, show error message
- [ ] validate if the JSON patch is valid or not, show error message
- [ ] if both valid, perform the edit, show success message
- [ ] show the diff after the succesful edit

ForgottenProgramme:
resolves #83 

This PR adds a new `patch` command allowing users to patch json blobs (from a file) to target recipe files. 

```
crm patch <json_blob_file_path> <target_recipe_file_path>
```

- [x] Add patch command to help message
- [x] validate if the file is a valid recipe or not, show error message
- [x] validate if the JSON patch is valid or not, show error message
- [x] if both valid, perform the edit, 
- [x] show success message
- [x] write tests
- [x] write more tests
- [x] parametrize the tests


tests/commands/test_patch.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+    runner = CliRunner()
+    fs.add_real_directory(TEST_FILES_PATH, read_only=False)  # type: ignore[attr-defined]
+
+    json_patch_path = TEST_FILES_PATH / "patch" / "correct_json_patch.json"
+    faulty_recipe_file_path = TEST_FILES_PATH / "patch" / "incorrect_recipe.yaml"
+
+    runner.invoke(patch, [str(json_patch_path), str(faulty_recipe_file_path)])
+    assert pytest.raises(IOError)
+    # assert result.exit_code == 5
```
schuylermartin45, line 34: Not sure if this is supposed to be commented out. We should test against the enum and not with a magic number

tests/commands/test_patch.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+    runner = CliRunner()
+    fs.add_real_directory(TEST_FILES_PATH, read_only=False)  # type: ignore[attr-defined]
+
+    faulty_json_patch_path = TEST_FILES_PATH / "patch" / "incorrect_json_patch.json"
+    recipe_file_path = TEST_FILES_PATH / "patch" / "correct_recipe.yaml"
+
+    result = runner.invoke(patch, [str(faulty_json_patch_path), str(recipe_file_path)])
+    assert pytest.raises(json.JSONDecodeError)
+    assert result.exit_code == 4
```
schuylermartin45, line 49: Again, we should use the enum and not a magic number.

tests/commands/test_patch.py
```diff
@@ -0,0 +1,63 @@
...
+
+    runner.invoke(patch, [str(json_patch_path), str(faulty_recipe_file_path)])
+    assert pytest.raises(IOError)
+    # assert result.exit_code == 5
+
+
+def test_faulty_json_patch_file(fs: pytest.Function) -> None:
+    """
+    Test to check the patch operation fails with a problematic json patch file
+    """
```
schuylermartin45, line 40: Should add a `:param fs:` docstring for the sake completeness/to help others who may write tests in the future. You can copy the comment from the `http_artifact_fetcher` tests.

I won't repeat this but this applies to every function in this test.

tests/commands/test_patch.py
```diff
@@ -0,0 +1,63 @@
...
+    Test to check the patch operation fails with a fauly recipe i.e a recipe which does not have the correct format etc
+    """
+    runner = CliRunner()
+    fs.add_real_directory(TEST_FILES_PATH, read_only=False)  # type: ignore[attr-defined]
+
+    json_patch_path = TEST_FILES_PATH / "patch" / "correct_json_patch.json"
+    faulty_recipe_file_path = TEST_FILES_PATH / "patch" / "incorrect_recipe.yaml"
+
+    runner.invoke(patch, [str(json_patch_path), str(faulty_recipe_file_path)])
+    assert pytest.raises(IOError)
```
schuylermartin45, line 33: Shouldn't this be a `with pytest.raises()`? Does this achieve the same thing? I actually don't know off the top of my head.

tests/commands/test_patch.py
```diff
@@ -0,0 +1,63 @@
...
+
+def test_faulty_recipe_file(fs: pytest.Function) -> None:
+    """
+    Test to check the patch operation fails with a fauly recipe i.e a recipe which does not have the correct format etc
+    """
+    runner = CliRunner()
+    fs.add_real_directory(TEST_FILES_PATH, read_only=False)  # type: ignore[attr-defined]
+
+    json_patch_path = TEST_FILES_PATH / "patch" / "correct_json_patch.json"
+    faulty_recipe_file_path = TEST_FILES_PATH / "patch" / "incorrect_recipe.yaml"
```
schuylermartin45, line 30: Should we consider parameterizing these? I know we only have 1 use case so far in these tests but it may be worth it to save time if we have to add another special case.

You can see an example of parameterizing with `fs`/fixtures in `test_http_artifact_fetcher.py`

tests/commands/test_patch.py
```diff
@@ -0,0 +1,63 @@
...
+
+    faulty_json_patch_path = TEST_FILES_PATH / "patch" / "incorrect_json_patch.json"
+    recipe_file_path = TEST_FILES_PATH / "patch" / "correct_recipe.yaml"
+
+    result = runner.invoke(patch, [str(faulty_json_patch_path), str(recipe_file_path)])
+    assert pytest.raises(json.JSONDecodeError)
+    assert result.exit_code == 4
+
+
+def test_healthy_recipe_and_json_patch_file(fs: pytest.Function) -> None:
```
schuylermartin45, line 52:
```suggestion
def test_patch_cli(fs: pytest.Function) -> None:
```
By the conventions of this project, we don't specify additional details in the test name when we test the "happy path". The comments should indicate specifics about the test case. If there are deviations from the normal or additional happy-paths to test, we mark those with a suffix to the test name like `test_foobar_additional_edge_case()`

tests/commands/test_patch.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def test_usage() -> None:
+    """
+    Smoke test that ensures rendering of the help menu
+    """
+    assert_cli_usage(patch)
+
+
+def test_faulty_recipe_file(fs: pytest.Function) -> None:
```
schuylermartin45, line 22:
```suggestion
def test_patch_cli_faulty_recipe_file(fs: pytest.Function) -> None:
```

This is more in-line with the naming scheme of the rest of the project. Refer to my other comment for more details. There's nothing necessarily wrong with the test name, but I want to try to be consistent. It also makes it easier to find similar tests if they share the same prefix.

tests/commands/test_patch.py
```diff
@@ -0,0 +1,63 @@
...
+
+    json_patch_path = TEST_FILES_PATH / "patch" / "correct_json_patch.json"
+    faulty_recipe_file_path = TEST_FILES_PATH / "patch" / "incorrect_recipe.yaml"
+
+    runner.invoke(patch, [str(json_patch_path), str(faulty_recipe_file_path)])
+    assert pytest.raises(IOError)
+    # assert result.exit_code == 5
+
+
+def test_faulty_json_patch_file(fs: pytest.Function) -> None:
```
schuylermartin45, line 37:
```suggestion
def test_patch_cli_faulty_patch_file(fs: pytest.Function) -> None:
```
Again, see previous comments.

tests/commands/test_patch.py
```diff
@@ -0,0 +1,63 @@
...
+    recipe_file_path = TEST_FILES_PATH / "patch" / "correct_recipe.yaml"
+
+    result = runner.invoke(patch, [str(faulty_json_patch_path), str(recipe_file_path)])
+    assert pytest.raises(json.JSONDecodeError)
+    assert result.exit_code == 4
+
+
+def test_healthy_recipe_and_json_patch_file(fs: pytest.Function) -> None:
+    """
+    Test to check the case when both the recipe file and the json patch file are healthy
```
schuylermartin45, line 54: I'd be more specific than to use the word "healthy" here. I would also use `JSON` in the documentation over `json`.

tests/commands/test_patch.py
```diff
@@ -0,0 +1,63 @@
...
+def test_usage() -> None:
+    """
+    Smoke test that ensures rendering of the help menu
+    """
+    assert_cli_usage(patch)
+
+
+def test_faulty_recipe_file(fs: pytest.Function) -> None:
+    """
+    Test to check the patch operation fails with a fauly recipe i.e a recipe which does not have the correct format etc
```
schuylermartin45, line 24:
```suggestion
    Test to check the patch operation fails with a bad recipe file (i.e a recipe which does not have the correct format, etc)
```
Nit picky, but I do like how you wrote this comment.

tests/commands/test_patch.py
```diff
@@ -0,0 +1,63 @@
...
+    faulty_recipe_file_path = TEST_FILES_PATH / "patch" / "incorrect_recipe.yaml"
+
+    runner.invoke(patch, [str(json_patch_path), str(faulty_recipe_file_path)])
+    assert pytest.raises(IOError)
+    # assert result.exit_code == 5
+
+
+def test_faulty_json_patch_file(fs: pytest.Function) -> None:
+    """
+    Test to check the patch operation fails with a problematic json patch file
```
schuylermartin45, line 39:
```suggestion
    Test to check that the patch operation fails with a invalid JSON patch file
```
Nitpicky, but I do like how this was written.

conda_recipe_manager/commands/patch.py
```diff
@@ -0,0 +1,71 @@
+"""
+:Description: CLI for patching JSON blobs to recipe files.
```
schuylermartin45, line 2:
```suggestion
:Description: CLI for patching JSON patch blobs to recipe files.
```

conda_recipe_manager/commands/patch.py
```diff
@@ -0,0 +1,71 @@
...
+from conda_recipe_manager.parser.exceptions import JsonPatchValidationException
+from conda_recipe_manager.parser.recipe_parser import RecipeParser
+from conda_recipe_manager.types import JsonPatchType
+
+
+def _pre_patch_validate(
+    json_patch_file_path: Path, recipe_file_path: Path
+) -> tuple[JsonPatchType | list[JsonPatchType], RecipeParser]:
+    """
+    Confirm that the json patch file and recipe file can be read and that the recipe parser object is created.
```
schuylermartin45, line 25: This is missing `:param:` and `:returns:` docs

conda_recipe_manager/commands/patch.py
```diff
@@ -0,0 +1,71 @@
...
+
+    return contents_json, recipe_parser
+
+
+@click.argument("recipe_file_path", type=click.Path(exists=True, path_type=Path))  # type: ignore[misc]
+@click.argument("json_patch_file_path", type=click.Path(exists=True, path_type=Path))  # type: ignore[misc]
+@click.command(short_help="Add JSON blobs to recipe files.")
+def patch(json_patch_file_path: Path, recipe_file_path: Path) -> None:
+    """
+    Patch JSON blobs to recipe files.
```
schuylermartin45, line 53:
```suggestion
    Patches recipe files with JSON patch blobs.
```
Going to be a bit nitpicky here since this is what is rendered when you run `crm patch --help`

conda_recipe_manager/commands/patch.py
```diff
@@ -0,0 +1,71 @@
...
+    except Exception:  # pylint: disable=broad-except
+        print_err("An error occurred while parsing the recipe file contents.")
+        sys.exit(ExitCode.PARSE_EXCEPTION)
+
+    return contents_json, recipe_parser
+
+
+@click.argument("recipe_file_path", type=click.Path(exists=True, path_type=Path))  # type: ignore[misc]
+@click.argument("json_patch_file_path", type=click.Path(exists=True, path_type=Path))  # type: ignore[misc]
+@click.command(short_help="Add JSON blobs to recipe files.")
```
schuylermartin45, line 50:
```suggestion
@click.command(short_help="Modify recipe files with JSON patch blobs.")
```
Going to be nitpicky here since this is what shows up when you run `crm --help`

tests/commands/test_patch.py
```diff
@@ -0,0 +1,63 @@
...
+from tests.smoke_testing import assert_cli_usage
+
+
+def test_usage() -> None:
+    """
+    Smoke test that ensures rendering of the help menu
+    """
+    assert_cli_usage(patch)
+
+
```
schuylermartin45, line 21: We should add tests to ensure we fail correctly when each file is given a path that doesn't exist.

tests/commands/test_patch.py
schuylermartin45: Actually the more I think about it, this `assert` shouldn't be here at all. We shouldn't be throwing in a CLI, we should be returning a POSIX-style error code.

conda_recipe_manager/commands/patch.py
```diff
@@ -0,0 +1,71 @@
...
+def patch(json_patch_file_path: Path, recipe_file_path: Path) -> None:
+    """
+    Patch JSON blobs to recipe files.
+    """
+    contents_json, recipe_parser = _pre_patch_validate(json_patch_file_path, recipe_file_path)
+
+    if not isinstance(contents_json, list):
+        contents_json = cast(list[JsonPatchType], list(contents_json))
```
schuylermartin45, line 58: An observation: If we didn't have such a sophisticated JSON schema validator in the recipe parser, we would have to perform some much more significant type checking to see if we were given an object or a list. But because we are relying on the schema validation, this is good!

tests/commands/test_patch.py
ForgottenProgramme: Ahh!

tests/commands/test_patch.py
dholth: Why not set the `fs` type to the correct `def test_faulty_json_patch_file(fs: pyfakefs.fake_filesystem.FakeFilesystem) -> None:` Will be picked up by docs tools and editor autocomplete without needing a special docstring.

schuylermartin45: I should have a PR up shortly that fixes this everywhere else.

schuylermartin45: @ForgottenProgramme I'm just going to do:
```
from pyfakefs.fake_filesystem import FakeFilesystem

def test_foo(fs: FakeFilesystem) -> None:
```

schuylermartin45: #128

tests/commands/test_patch.py
ForgottenProgramme: but click does that for us anyways.

schuylermartin45: I suppose this is where the line between `unit` and `integration` testing gets a bit blurry. But it is a pretty easy thing to test. 

Here's a hypothetical that could happen/I've seen before (not specifically with `click` but with other libraries):

There could be some day in the future where we decide not to use `click` any more. In that event, we replace `click`. If there is no test and the new system doesn't work like `click` does, we could have a regression. Best to cover our butts now than have it blow up in our face.

ForgottenProgramme:
I think the tests added in test_patch.py should pass all the checks/integration tests.

ForgottenProgramme:
@schuylermartin45, I am trying to test for the case when the patch file only contains a single patch instead of a list of patches. In this case the code in `patch.py` should first put the patch in a list. But looks like that's not happening because 
1. the test `test_patch_cli_single_patch` is failing
2. When I run `crm patch single_patch.json recipe.yaml` , I get the following: 
`The patch provided did not follow the expected schema.`  which comes from line 75 in `patch.py`

tests/test_aux_files/patch/bad_recipe_files/missing_colon.yaml
```diff
@@ -0,0 +1,2 @@
+package
+  name: {{ name|lower }}  # [unix]
```
schuylermartin45, line 2: Good job on this one. I don't know why I didn't think of this kind of recipe corruption sooner.

schuylermartin45 (approved):

PR #107, opened 2024-08-27
Adds minor QOL improvements to `crm graph`

schuylermartin45:
- Adds a missing help menu string
- `graph` now uses the same `ExitCode` enumerations that the other commands use.

### DIFF
--- a/conda_recipe_manager/commands/graph.py
+++ b/conda_recipe_manager/commands/graph.py
@@ -14,12 +14,13 @@ from typing import Final
 import click
 
 from conda_recipe_manager.commands.utils.print import print_err
+from conda_recipe_manager.commands.utils.types import ExitCode
 from conda_recipe_manager.grapher.recipe_graph import PackageStats, RecipeGraph
 from conda_recipe_manager.grapher.recipe_graph_from_disk import RecipeGraphFromDisk
 from conda_recipe_manager.grapher.types import GraphDirection, GraphType, PackageStatsEncoder
 
 
-@click.command(short_help="")
+@click.command(short_help="Interactive CLI for examining recipe dependency graphs.")
 @click.argument("path", type=click.Path(exists=True, path_type=Path, file_okay=False))  # type: ignore[misc]
 def graph(path: Path) -> None:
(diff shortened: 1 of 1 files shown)


conda_recipe_manager/commands/graph.py
```diff
@@ -34,8 +35,7 @@ def graph(path: Path) -> None:
     recipe_graph: Final[RecipeGraph] = RecipeGraphFromDisk(path)
     if not recipe_graph:
         print_err(f"The path provided does not contain any recipe files: {path}")
-        # TODO use enum scheme
-        sys.exit(1)
+        sys.exit(ExitCode.CLICK_USAGE)
```
jezdez, line 38: Any reason to not use click's native abort system? Something like `ctx.fail(f"The path provided does not contain any recipe files: {path}")`

jezdez (approved):

ForgottenProgramme (approved):

conda_recipe_manager/commands/graph.py
schuylermartin45: It doesn't appear that that function allows you to control the error codes. I have have this standardized enumeration to ensure that all the CRM commands emit similar error codes and that I control what the error codes are.

Trying to go for that UNIX design philosophy and allow other programs to rely on error codes having a particular meaning.

jezdez: Right, that's a good policy, I like it! For custom exit codes click has built-in error handling via Python exceptions: https://click.palletsprojects.com/en/8.1.x/exceptions/#where-are-errors-handled

Here's the code that does it: https://github.com/pallets/click/blob/99015e1456b6952054fda8312b40ea646737168d/src/click/core.py#L1320-L1324. 

[`ClickException` has a bunch of subclasses that you could reuse, if they match your use case](https://github.com/pallets/click/blob/99015e1456b6952054fda8312b40ea646737168d/src/click/exceptions.py#L25). The key part is that for custom exit codes, simply subclass and define an `exit_code` class attribute with your values.

PR #114, opened 2024-09-06
Migrates CRM to use yaml.CSafeLoader over yaml.safe_load()

schuylermartin45:
TODO


dholth:
Uses it transparently?

PR #115, opened 2024-09-06
Migrates CRM to use yaml.CSafeLoader over yaml.safe_load()

schuylermartin45:


schuylermartin45:
Update: this makes a much more noticeable reduction in the time it takes to complete the GHA tests. It is harder to notice the difference locally when you have an Apple Silicon number of CPU cores.

conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -97,10 +97,10 @@ def _v1_sub_jinja() -> None:
         # then we fall back to performing JINJA substitutions.
         try:
             try:
-                output = cast(JsonType, yaml.safe_load(s))
+                output = yaml.load(s, Loader=yaml.CSafeLoader)
```
JeanChristopheMorinPerso, line 100: You might want to do something like
```python
try:
    from yaml import CSafeLoader as SafeLoader
except ImportError:
    from yaml import SafeLoader
```
to support cases where the C bindings are not available.

schuylermartin45:
> In conda we've switched over completely to ruamel.yaml... mainly because it is YAML 1.2 instead of 1.1, and has among other things "Only true and false strings are parsed as booleans (including True and TRUE); y, yes, on, and their negative counterparts are parsed as strings."

Last I looked, switching is not a fast drop-in replacement. I think we'll have to look into it further to decide if it is worth the change. Personally, I have not had the best developer experience with ruamel.

beeankha (approved):

PR #117, opened 2024-09-10
Adds file mocking module

schuylermartin45:
This module provides tools to mock writing to a single file.

This work is based on the suggestions from this StackOverflow post: https://stackoverflow.com/questions/70861340/how-to-check-what-was-written-to-a-mock-open-fake-file

Here is some pseudo code to illustrate how to use this work:

```
from unittest import mock
from tests.file_mocking import MockWriter

mock_opener, mock_writer = MockWriter.setup_file_mockers("Mocked file contents")

with mock.patch('<function that calls to write()>', mock_opener) as f:
    <call to function that writes to file>
    assert str(mock_writer) == expected

```

### DIFF
--- a/tests/file_mocking.py
+++ b/tests/file_mocking.py
@@ -74,7 +74,7 @@ class MockWriter:
     def setup_file_mockers(mocked_data: str) -> tuple[MockOpener, MockWriter]:
         """
         Convenience function that sets up a mocked file opener and a `MockWriter` instance to be used for testing that
-        a function writes the expected `mocked_data` to a file.
+        writes the expected `mocked_data` to a file.
 
         :param mocked_data: Data that should be contained in the file that is being written to.
         :returns: The mock opener and mock writer instances configured to emulate a file that should contain
--- a/tests/file_mocking.py
+++ b/tests/file_mocking.py
@@ -1,0 +1,86 @@
+"""
+:Description: Test utility module that provides file mocking tools.
+"""
+
+from __future__ import annotations
+
+from typing import Callable, cast
+from unittest import mock
+
+
+class _Writeable:
+    """
+    Dummy class used to provide type annotations to `return_value` in the object returned by `mock.mock_open()`.
+    """
+
+    write: Callable[[str], None]
+
+
+class MockOpener:
+    """
+    Dummy class used to provide type annotations to the object returned by `mock.mock_open()`.
+    """
+
+    return_value: _Writeable
+
+
+class MockWriter:
+    """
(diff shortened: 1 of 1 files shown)


dholth:
Why not https://docs.python.org/3/library/io.html#binary-i-o or StringIO?

schuylermartin45:
> Why not https://docs.python.org/3/library/io.html#binary-i-o or StringIO?

I don't anticipate that we will be using this enough times or with large enough files that we will notice much of a performance gain. But given the writer class hides its data, we can iterate on that in the future.

dholth:
It would be usual to return a BytesIO or StringIO instead of the Writable class. Or a regular file under tmp_path. https://docs.pytest.org/en/stable/how-to/tmp_path.html

schuylermartin45:
> It would be usual to return a BytesIO or StringIO instead of the Writable class. Or a regular file under tmp_path. https://docs.pytest.org/en/stable/how-to/tmp_path.html

I am extremely against writing to disk for unit tests. That introduces a resource that is not easily controlled.

Here are my biggest issues:
- The tests could fail if the file system is full or you are on a platform where the file system is not writeable.
- If there is an issue with the file system, the unit tests can fail for a reason other than there is an issue with the unit of software being tested.
- Writing to disk introduces the possibility of resource contention if the file being written to is required by another test. If all files in the test suite are only read by tests, you can safely parallelize the tests. I've seen this happen enough times and would prefer to avoid the possibility completely.

I see writing to the file system as a resource out of your control that must be mocked. You wouldn't want a unit test to rely on a network resource (REST API, web page, etc) for the same reasons.

dholth:
Main point is that there is an implementation of MockWriter in the standard library's io module. If you happen to need the filesystem, "You can use the tmp_path fixture which will provide a temporary directory unique to each test function." according to pytest's documentation.

tests/file_mocking.py
```diff
@@ -0,0 +1,86 @@
...
+        """
+        if not isinstance(o, MockWriter):
+            return False
+        return str(self) == str(o)
+
+    @staticmethod
+    def setup_file_mockers(mocked_data: str) -> tuple[MockOpener, MockWriter]:
+        """
+        Convenience function that sets up a mocked file opener and a `MockWriter` instance to be used for testing that
+        a function writes the expected `mocked_data` to a file.
```
beeankha, line 77: Minor wording nitpick

```suggestion
        writes the expected `mocked_data` to a file.
```

beeankha (approved):

PR #120, opened 2024-09-13
Deprecates MockWriter class in favor of `pyfakefs`

schuylermartin45:
This work is no longer necessary now that we are using `pyfakefs`. Best delete it now before it is used and we have two competing solutions to the same problem.


dholth (approved):
Solid removal.

PR #122, opened 2024-09-17
Implements `GitArtifactFetcher`

schuylermartin45:
Completes the previously templated-out `GitArtifactFetcher` class.

This PR also contains some minor refactoring on the previous Artifact Fetcher work.

As a simple wrapper around `GitPython`, it does not seem worth our current time to develop significant unit testing of the `GitArtifactFetcher` class. We would have to replace so much of the underlying infrastructure (i.e. `GitPython`) with mocking, we would only be unit testing a few lines of the actual class.

I think in the future, we should develop an integration test instead.


beeankha (approved):

conda_recipe_manager/fetcher/base_artifact_fetcher.py
```diff
@@ -55,3 +69,9 @@ def get_path_to_source_code(self) -> Path:
 
         :raises FetchRequiredError: If a call to `fetch()` is required before using this function.
         """
+
+    def apply_patches(self) -> None:
+        """
+        TODO Flush this mechanism out. It looks like the same mechanism is used for http and git sources(?)
```
dholth, line 75: flesh

conda_recipe_manager/fetcher/git_artifact_fetcher.py
```diff
@@ -4,40 +4,120 @@
...
         """
-        pass
+        match self._is_remote:
+            case True:
+                return Repo.clone_from(self._git_target.url, self._temp_dir_path)  # type: ignore[misc]
+            case False:
+                shutil.copytree(self._git_target.url, self._temp_dir_path)
+                return Repo(self._temp_dir_path)
+
+    def _resolve_checkout_target(self) -> Optional[str]:
```
dholth, line 71: `Optional[str]` can be spelled `str | None`

schuylermartin45: I know, but this is arguably more human readable and in-line with nomenclature used by other languages .

marcelotrevisani: well, no... it boils down to preference other than human-readable.
I can argue that ``str | None`` is more readable since, I know the two types of the return when you compare to ``Optional[str]``, well, what is ``Optional`` in python? that is more implicit.

and which programming languages? typescript uses ``str | null``, rust you can use ``Option<T>`` but usually put that in a result.
swift and C# uses ``str?``

schuylermartin45: I have spent a significant amount of time with modern C++ and that uses [optional](https://en.cppreference.com/w/cpp/utility/optional). Maybe I'm biased from that. Apparently [Java does as well](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html) and I would argue `rust`'s `Option` is just two letters away from being `Option`-`al`.

Regardless, I don't find style arguments worth anyone's time. There are much bigger problems to solve in our community.

This is the lane we picked in this project and I choose to stay in that lane unless I hear someone claim there is a significant practical difference.

PR #124, opened 2024-09-18
Completes the first iteration of `from_recipe()`

schuylermartin45:
- Adds unit testing
- Includes a new `get_recipe_name()` convenience function to standardize the answering of the "what to call a recipe" problem
- Includes a simple `optional_str()` utility that forces string-evaluation unless a value is `None`.

### DIFF
--- a/conda_recipe_manager/fetcher/artifact_fetcher.py
+++ b/conda_recipe_manager/fetcher/artifact_fetcher.py
@@ -1,33 +1,100 @@
 """
-:Description: TODO
+:Description: Module that provides general Artifact Fetching utilities and factory methods.
 """
 
 from __future__ import annotations
 
-from typing import Final, cast
+from typing import cast
 
 from conda_recipe_manager.fetcher.base_artifact_fetcher import BaseArtifactFetcher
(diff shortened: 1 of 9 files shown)


conda_recipe_manager/fetcher/artifact_fetcher.py
```diff
@@ -1,33 +1,100 @@
...
+                case "git_url":
+                    return "git"
+                case "git_branch":
+                    return "branch"
+                case "git_tag":
+                    return "tag"
+                case "git_rev":
+                    return "rev"
+                case _:  # If this case happens, a developer made a typo.
+                    return ""
```
schuylermartin45, line 42: I've gone back and forth on this line a few times. I think my current thinking is that the risk of this silently failing is extremely low and should be caught in any sufficient unit test coverage.

schuylermartin45: I've changed my mind. It is within reason that this could throw a `FetchUnsupportedError()`. It would still be the fault of the developer for putting in a bad string, but a typo should be caught and be made obvious in a unit test that throws.

beeankha (approved):

PR #153, opened 2024-09-23
Addresses nuance with pyfakefs and pathlib

### LINKED ISSUE #151: `TEST_FILES_PATH` uses a `Path` object outside of `pyfakefs`
@ForgottenProgramme found this: https://github.com/conda-incubator/conda-recipe-manager/actions/runs/10996671650/job/30530378171?pr=106#step:4:66

Tests seem to fail sporadically in our GitHub Actions.

I believe the issue is caused by `TEST_FILES_PATH` being defined as a `Path` object globally. We need to use a string version of the path when dealing with `pyfakefs` to avoid a bootstrapping issue.

Docs on the matter:
- https://github.com/pytest-dev/pyfakefs/issues/553
- https://pytest-pyfakefs.readthedocs.io/en/latest/troubleshooting.html#pathlib-path-objects-created-outside-of-tests

schuylermartin45:
Fixes #151 by deprecating the global `TEST_FILE_PATHS` constant in favor of a function that generates the same `Path` object.

Although this is a more expensive approach, it does mean that every test has the conveniences offered by `pathlib` without having to worry about how the `Path` object is constructed. 

In other words:
Calling `get_test_path()` will create an object instance that refers to either the real or the fake file system, depending on the inclusion of the `fs` fixture.

### DIFF
--- a/tests/commands/test_convert.py
+++ b/tests/commands/test_convert.py
@@ -5,7 +5,7 @@
 from click.testing import CliRunner
 
 from conda_recipe_manager.commands.convert import convert
-from tests.file_loading import TEST_FILES_PATH, load_file
+from tests.file_loading import get_test_path, load_file
 from tests.smoke_testing import assert_cli_usage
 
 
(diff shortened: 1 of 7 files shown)


schuylermartin45:
@ForgottenProgramme you'll need to upgrade your tests on your branch to use this new mechanism.

beeankha (approved):

PR #154, opened 2024-09-23
Fixes #121

### LINKED ISSUE #121: Invalid "script: commands" in tests of generated recipe
Given the following recipe:

```
{% set name = "elmitec" %}
{% set version = "1.4.2.dev3+g8c30ab1.d20240913" %}

package:
  name: {{ name|lower }}
  version: {{ version }}

source:
  path: file:///build/dist/elmitec-1.4.2.dev3+g8c30ab1.d20240913.tar.gz
  sha256: c82e5b13a77550056be9bb8917dc3999786385e3118a4a01d2cb5183e8360c2f

build:
  noarch: python
  script: python -m pip install . -vv --no-deps --no-build-isolation
  number: 0

requirements:
  host:
    - python
    - setuptools >=45
    - wheel
    - setuptools-scm >=6.2,<8
    - pip
  run:
    - python
    - numpy

test:
  imports:
    - elmitec
  commands:
    - pip check
  requires:
    - pip

about:
  home: http://www.maxiv.lu.se
  summary: An collection of API to manage the Elmitec Uview program and the Leem microscope
  license: GPL-3.0
  license_file: LICENSE.txt
```

`crm convert` outputs a line `script: commands` that is invalid.
And my understanding is that there is no need to put `pip` in the tests run requirements, when `pip_check` is true.

```
$ crm convert meta.yaml
schema_version: 1

context:
  name: elmitec
  version: 1.4.2.dev3+g8c30ab1.d20240913

package:
  name: ${{ name|lower }}
  version: ${{ version }}

source:
  sha256: c82e5b13a77550056be9bb8917dc3999786385e3118a4a01d2cb5183e8360c2f
  path: file:///build/dist/elmitec-1.4.2.dev3+g8c30ab1.d20240913.tar.gz

build:
  number: 0
  noarch: python
  script: python -m pip install . -vv --no-deps --no-build-isolation

requirements:
  host:
    - python
    - setuptools >=45
    - wheel
    - setuptools-scm >=6.2,<8
    - pip
  run:
    - python
    - numpy

tests:
  - python:
      imports:
        - elmitec
      pip_check: true
  - requirements:
      run:
        - pip
    script: commands

about:
  summary: An collection of API to manage the Elmitec Uview program and the Leem microscope
  license: GPL-3.0
  license_file: LICENSE.txt
  homepage: http://www.maxiv.lu.se
```

schuylermartin45:
Fixes #121 by adding a new logic check. Also adds a regression test to catch this edge case.

### DIFF
--- a/conda_recipe_manager/parser/recipe_parser_convert.py
+++ b/conda_recipe_manager/parser/recipe_parser_convert.py
@@ -590,6 +590,12 @@ class RecipeParserConvert(RecipeParser):
             self._patch_and_log({"op": "remove", "path": RecipeParser.append_to_path(test_path, f"/commands/{i}")})
             pip_check = True
             break
+
+        # Edge-case: Remove `script` if pip-check was the only command present
+        if pip_check and len(commands) == 1:
+            # TODO calc path
+            self._patch_and_log({"op": "remove", "path": RecipeParser.append_to_path(test_path, "/script")})
+
         self._patch_add_missing_path(test_path, "/python")
         self._patch_and_log(
             {"op": "add", "path": RecipeParser.append_to_path(test_path, "/python/pip_check"), "value": pip_check}
--- a/conda_recipe_manager/parser/recipe_parser_convert.py
+++ b/conda_recipe_manager/parser/recipe_parser_convert.py
@@ -580,21 +580,27 @@ class RecipeParserConvert(RecipeParser):
         ):
             return
 
-        commands = cast(list[str], self._v1_recipe.get_value(RecipeParser.append_to_path(test_path, "/commands"), []))
+        commands_path: Final[str] = RecipeParser.append_to_path(test_path, "/commands")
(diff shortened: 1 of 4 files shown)


schuylermartin45:
Interesting, the `script` section is required and this current fix actually makes compatibility much worse with `rattler-build`.   
I'll have to have a think on this.

But good job integration tests for catching something ahead of time!

schuylermartin45:
After some thinking/rubber-ducking: In this case, the entire `requirements/script` blob doesn't actually need to exist and should be removed entirely. Only the `python` test blob needs to remain.

schuylermartin45:
In the latest commit: the integration testing gained 5 more successful recipe conversions against the `rattler-build` dry-run. Small potatoes, but we definitely fixed the regression. I'll take that win.

beeankha (approved):

PR #176, opened 2024-10-02
Update `convert` to properly handle complex Jinja expressions

beeankha:
Resolves https://github.com/conda-incubator/conda-recipe-manager/issues/166


conda_recipe_manager/parser/_types.py
```diff
@@ -116,7 +116,7 @@ class Regex:
     """
 
     # Pattern to detect Jinja variable names and functions
-    _JINJA_VAR_FUNCTION_PATTERN: Final[str] = r"[a-zA-Z_][a-zA-Z0-9_\|\'\"\(\)\[\]\, =\.\-]*"
+    _JINJA_VAR_FUNCTION_PATTERN: Final[str] = r"[a-zA-Z_][a-zA-Z0-9_\|\'\"\(\)\[\]\, =\.\-~]*"
```
schuylermartin45, line 119: I actually don't see `~` in the proposed CEP-17, so I'm not sure if it will be supported.
I also need to figure out what the hell that actually does in JINJA, lol.

schuylermartin45: Okay so from context of the original post, I think the `~` is tied to semantic versioning. Now why that's valid outside of the string argument in the `pin_subpackage()` example, I'm less sure.

conda_recipe_manager/parser/recipe_parser_convert.py
```diff
@@ -171,6 +172,22 @@ def _upgrade_jinja_to_context_obj(self) -> None:
...
+                if isinstance(value, str) and any(pattern.search(value) for pattern in Regex.V0_UNSUPPORTED_JINJA)
+            ]
+            if complex_jinja:
+                complex_jinja_pattern = [re.sub(r"\\(.)", r"\1", regex.pattern) for regex in Regex.V0_UNSUPPORTED_JINJA]
+                complex_jinja_display = ", ".join(complex_jinja)
+                complex_jinja_pattern_display = ", ".join(complex_jinja_pattern)
+                self._msg_tbl.add_message(
+                    MessageCategory.WARNING,
+                    f"Complex Jinja expressions detected in key(s): {complex_jinja_display}\n"
+                    f"The following syntax cannot be automatically converted: {complex_jinja_pattern_display}",
```
schuylermartin45, line 188: I would simplify this into one warning line. I don't think we have to show the regex pattern, just the key associated with the failure.

schuylermartin45:
If you compare with the last integration test with main, this fixes compatibility with 6 recipes. Nice!

PR #196, opened 2024-10-11
Test the `MessageTable` class

### LINKED ISSUE #177: Add missing unit tests for `MessageTable` class
The `MessageTable` class does not currently have unit tests associated with it. Currently the class can be found in the top-level `conda_recipe_manager.types` module.

To match the current convention, the new test file should be added as `tests/test_types.py`. There already exists a `test_types.py` file, so we may need to add `__init__.py` files first, per [this documentation](https://docs.pytest.org/en/latest/explanation/pythonpath.html). Issue #178 tracks this.

beeankha:
Resolves #177 

The new unit test aims to cover all of the methods contained in the [`MessageTable` class in `types.py`](https://github.com/conda-incubator/conda-recipe-manager/blob/main/conda_recipe_manager/types.py#L84-L159), which includes `add_message()`, `get_messages()`, `get_message_count()`, `get_totals_message()`, and `clear_messages()`.

### DIFF
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -4,27 +4,62 @@
 
 from __future__ import annotations
 
+import pytest
+
 from conda_recipe_manager.types import MessageCategory, MessageTable
 
 
-def test_message_table_add_retrieve_and_clear() -> None:
+@pytest.mark.parametrize(
+    "message,content,output",
+    [
+        (MessageCategory.WARNING, "Test warning message", "0 errors and 1 warning were found."),
+        (MessageCategory.ERROR, "Test error message", "1 error and 0 warnings were found."),
+    ],
+)
+def test_message_table_add(message: MessageCategory, content: str, output: str) -> None:
+    """
+    Tests the addition of messages to the message table and checks output of warnings and errors contained
+    in the messaging object.
+
+    param message: Message category
+    param content: Message content
+    param output: Expected output
+    """
     message_table = MessageTable()
-    message_table.add_message(MessageCategory.ERROR, "Test error message")
-    message_table.add_message(MessageCategory.WARNING, "Test warning message")
+    message_table.add_message(message, content)
+    assert message_table.get_totals_message() == output
 
-    assert message_table.get_totals_message() == "1 error and 1 warning were found."
 
-    assert message_table.get_message_count(MessageCategory.EXCEPTION) == 0
-    assert message_table.get_message_count(MessageCategory.ERROR) == 1
-    assert message_table.get_message_count(MessageCategory.WARNING) == 1
+@pytest.mark.parametrize(
+    "message,content,count",
+    [
+        (MessageCategory.EXCEPTION, "Test exception message", 1),
+        (MessageCategory.WARNING, "Test warning message", 1),
+        (MessageCategory.ERROR, "Test error message", 1),
+    ],
+)
+def test_message_get_count(message: MessageCategory, content: str, count: int) -> None:
+    """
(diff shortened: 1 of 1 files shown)


tests/test_types.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def test_message_table_add_retrieve_and_clear() -> None:
+    message_table = MessageTable()
+    message_table.add_message(MessageCategory.ERROR, "Test error message")
+    message_table.add_message(MessageCategory.WARNING, "Test warning message")
+
+    assert message_table.get_totals_message() == "1 error and 1 warning were found."
+
+    assert message_table.get_message_count(MessageCategory.EXCEPTION) == 0
```
schuylermartin45, line 17: I would prefer we have lots of little tests over one big test that checks many things. Small tests can be highly parallelized and narrow the search space of a potential problem.

I would also challenge you to see if you can parameterize these tests so we can build more tests faster.

schuylermartin45 (changes requested):
See comments above.

tests/test_types.py
```diff
@@ -0,0 +1,68 @@
...
+from conda_recipe_manager.types import MessageCategory, MessageTable
+
+
+@pytest.mark.parametrize(
+    "message,content,output",
+    [
+        (MessageCategory.WARNING, "Test warning message", "0 errors and 1 warning were found."),
+        (MessageCategory.ERROR, "Test error message", "1 error and 0 warnings were found."),
+    ],
+)
```
beeankha, line 18: @schuylermartin45 is there an effective way that I could add both a warning AND an error via parametrization and get the `"1 error and 1 warning were found."` output? I tried several things but kept getting a `TypeError` with too many arguments passed to `MessageTable.add_message()`.

schuylermartin45: Although I normally advise not to put for loops in tests, in this case, I think you could combine `message` and `content` into a list of tuples to make it easier to develop test cases:
```

([(MessageCategory.WARNING, "Test warning message")], "0 errors and 1 warning were found."),
([(MessageCategory.ERROR, "Test error message"), (MessageCategory.WARNING, "Test warning message")], "1 error and 1 warning were found."),
```

tests/test_types.py
```diff
@@ -0,0 +1,154 @@
...
+            ("0 errors and 0 warnings were found."),
+        ),
+    ],
+)
+def test_message_table_add(messages_and_contents: list[tuple[MessageCategory, str]], expected: str) -> None:
+    """
+    Tests the addition of messages to the message table and checks the expected output of warnings and errors
+    contained in the messaging object.
+
+    param messages_and_contents: Message category and message content
```
schuylermartin45, line 63:
```suggestion
    :param messages_and_contents: Message category and message content
```

schuylermartin45: Actually all of these are like this....

schuylermartin45: I have a pylint rule to enforce doc strings but it seems to be flaky/inconsistent.

tests/test_types.py
```diff
@@ -0,0 +1,154 @@
...
+    param count: Expected count
+    """
+    message_table = MessageTable()
+    for category, content in messages_and_contents:
+        message_table.add_message(category, content)
+
+    # Check that there are actually warning and error messages being registered and counted before clearing them
+    for category, counts in expected_count.items():
+        assert message_table.get_message_count(category) == counts
```
schuylermartin45, line 150: I feel like I could go either way on if this is necessary to check. I'm okay leaving it for now though.

schuylermartin45 (approved):

PR #199, opened 2024-10-11
Fixes #197 and adds regression tests. Also fixes #190

### LINKED ISSUE #190: Complex JINJA statements are not being evaluated in strings.
There are two commented-out test case in the PR for #186 (in `test_recipe_reader.py`):
```
("sub_vars.yaml", "/requirements/run_constrained/15", True, "foo > 42"),
...
("v1_format/v1_sub_vars.yaml", "/requirements/run_constraints/15", True, "foo > 42"),
```

Complex JINJA expressions that are contained in strings currently do not get evaluated by `RecipeReader::get_value(.., sub_vars=True)`:
```
foo > {{ '4' + "2" }}
```

### LINKED ISSUE #197: over quoting jinja2 items
**What happened?**

It appears that CRM is overquoting some jinja2 items. See this bug report comment on smithy for the error and steps to reproduce: https://github.com/conda-forge/conda-smithy/issues/2085#issuecomment-2407734898

**To Reproduce**

**Additional Details (please complete the following information):**
 - OS: [e.g. MacOS M1, Windows]
 - Conda Recipe Manager version [e.g. `0.2.0`]
 - Any other context about the problem that was found.

schuylermartin45:
Fixes #197 and adds regression tests. Also fixes #190

We were over-quoting some V0 JINJA strings by accident. This lead to the improper parsing of some YAML lines.

### DIFF
--- a/conda_recipe_manager/parser/_utils.py
+++ b/conda_recipe_manager/parser/_utils.py
@@ -132,7 +132,9 @@ def quote_special_strings(s: str, multiline_variant: MultilineVariant = Multilin
     # without substitution markers (like `match()`)
     if (
         multiline_variant != MultilineVariant.NONE
-        or Regex.JINJA_V0_SUB.match(s)
+        # We check the entire string for JINJA statements to avoid quoting valid YAML strings like:
+        # `- ${{ compiler('rust') }} >=1.65.0` and `foo > {{ '4' + "2" }}`.
+        or cast(list[str], Regex.JINJA_V0_SUB.findall(s))
         or Regex.JINJA_FUNCTION_MATCH.search(s)
     ):
         return s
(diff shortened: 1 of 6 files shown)


schuylermartin45:
I want to verify that this does not cause a regression in the conversion integration tests before merging.

schuylermartin45:
Overall we decrease our compatibility by 3 recipe files out of 1300. However, if you look closer at the individual tests, there are some compatibility improvements.

This looks like one of those cases where one problem silently masked other issues. I'm okay with that trade-off, especially given this bug was found by our conda-forge users.

schuylermartin45:
Didn't think this would happen before the weekend, but I also nailed down #190 and will include that fix in this PR since the two are very similar issues.

beeankha (approved):

PR #264, opened 2024-12-04
260 Parallelize multiple source fetch requests

schuylermartin45:
We now use futures to download multiple source files concurrently.

Also introduces a `--dry-run` mode that dumps the recipe file to STDOUT instead of overwriting the recipe file.

~~Putting this PR up as a Draft until I can figure out why some of the `bump_recipe_cli` tests are significantly slower after this change (causing the tests to take 4 seconds to 35 seconds).~~ That issue is now tracked in #265 as it appears to have no known user impact and is tied to the test environment.

### DIFF
--- a/conda_recipe_manager/commands/bump_recipe.py
+++ b/conda_recipe_manager/commands/bump_recipe.py
@@ -335,6 +335,12 @@ def _update_sha256(recipe_parser: RecipeParser, retry_interval: float) -> None:
     help="Bump the build number by 1.",
 )
 @click.option(
+    "-d",
+    "--dry-run",
+    is_flag=True,
+    help="Performs a dry-run operation that prints the recipe to STDOUT and does not save to the recipe file.",
+)
+@click.option(
     "-t",
     "--target-version",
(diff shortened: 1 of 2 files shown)


schuylermartin45:
The `test_bump_recipe_cli()` tests that reference
- `gsm-amzn2-aarch64_version_bump.yaml`
- `libprotobuf_version_bump.yaml`
cause the slow down. Each of these cases takes ~30 seconds to complete. They both have multiple sources but they are fully mocked.

schuylermartin45:
The slow down goes away if I remove `pyfakefs` as a test fixture for `test_bump_recipe_cli()`

schuylermartin45:
Upgrading to `pyfakefs` to `5.7.2` seems to help, but it could be transient. Some runs take 4 seconds, other runs still take 30+ seconds.

schuylermartin45:
Can't reproduce this slow down in a real (non-test) environment. Will spin-off this issue to another ticket, as it does not affect users.

PR #271, opened 2024-12-13
properly convert env interactions into context

AaronOpfer:
This input:

```yaml
{% set a = environ["a"] %}
{% set b = environ.get("b", "0") %}

package:
  name: abc
  version: 0
```

on `main` emits this recipe that needs manual correction:

```yaml
schema_version: 1

context:
  a: "\"env.get(\"a\")\""
  b: "\"environ | get(\"b\", \"0\")\""

package:
  name: abc
  version: 0
```

but after this PR, it will instead emit:
```yaml
schema_version: 1

context:
  a: ${{env.get("a")}}
  b: ${{env.get("b", default="0")}}

package:
  name: abc
  version: 0
```
which does not need any further editing.

This patch feels a little hacky, and I would especially like to immortalize my test case. I'm happy to accept feedback on how to go about making this patchset less hacky and immortalize my test case.

Thanks for your time.


travishathaway:
@conda-bot check

AaronOpfer:
CLA is sorted now. It looks like CI is blocked on that.

schuylermartin45:
It looks like the code formatter, linter, and static analyzer checks failed.

Let me know if you need assistance setting that up locally. Any feedback on the README or other documentation is very welcome. It would be nice to know if there is room for improvement!

conda_recipe_manager/parser/_types.py
```diff
@@ -132,6 +132,11 @@ class Regex:
     # Finds `environ[]` used by a some recipe files. Requires a whitespace character to prevent matches with
     # `os.environ[]`, which is very rare.
     PRE_PROCESS_ENVIRON: Final[re.Pattern[str]] = re.compile(r"\s+environ\[(\"|')(.*)(\"|')\]")
+
+    # Finds `environ.get` which is used in a closed source community of which the author of this comment
+    # participates in.
```
schuylermartin45, line 137: I really, really appreciate having context comments. I'll just amend this to include the PR.
```suggestion
    # participates in. See Issue #271 for more details.
```

conda_recipe_manager/parser/recipe_parser_convert.py
```diff
@@ -132,7 +132,7 @@ def _upgrade_jinja_to_context_obj(self) -> None:
                 self._msg_tbl.add_message(MessageCategory.WARNING, f"The variable `{name}` is an unsupported type.")
                 continue
             # Function calls need to preserve JINJA escaping or else they turn into unevaluated strings.
-            if isinstance(value, str) and search_any_regex(Regex.JINJA_FUNCTIONS_SET, value):
+            if isinstance(value, str) and (search_any_regex(Regex.JINJA_FUNCTIONS_SET, value) or value.startswith("env.get")):
```
schuylermartin45: Part of me wonders if `startswith` should be replaced with a regex to be more versatile, in case a direct string compare fails on whitespace or formatting. I think at this point `value` should be stripped, so maybe I'm over thinking this.

At the very least, I'll add this breadcrumb comment:
```suggestion
            # See Issue #271 for more details on the `env.get` case.
            if isinstance(value, str) and (search_any_regex(Regex.JINJA_FUNCTIONS_SET, value) or value.startswith("env.get")):
```

conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -704,7 +704,7 @@ def render(self) -> str:
         if self._schema_version == SchemaVersion.V0:
             for key, val in self._vars_tbl.items():
                 # Double quote strings
-                if isinstance(val, str):
+                if isinstance(val, str) and not val.startswith('env.get('):
```
schuylermartin45: Any reason why this line checks against the `(` and the other check doesn't? If we don't make this a regex, maybe we should make the string a constant?
```suggestion
                # See Issue #271 for more details on the `env.get` case.
                if isinstance(val, str) and not val.startswith('env.get('):
```

conda_recipe_manager/parser/recipe_parser_convert.py
```diff
@@ -738,6 +740,16 @@ def pre_process_recipe_text(content: str) -> str:
                     f"env.get({quote_char}{key}{quote_char})",
                 )
             )
+
+        for groups in cast(list[tuple[str, ...]], Regex.PRE_PROCESS_ENVIRON_GET.findall(content)):
+            key_quote_char, key, _, default_quote_char, default, _ = groups
```
schuylermartin45, line 745: Arguably, I should use this expansion form over indexing in the other transformations :)

AaronOpfer:
These are my notes in no particular order:
-  The `make dev` fails on up-to-date tooling if the `conda-recipe-manager` environment doesn't already exist as the conda command exits non-zero. I had to create a token env first to satisfy the makefile.
- The makefile doesn't work on my ancient CentOS7 `make` version 3.82. I needed to source `make` from `conda-forge` before I could proceed.
- Various tooling seems to be straddling `defaults` and `conda-forge` which is a no-no.
- Attempting to use the assisted build inside my company's conda configuration was impossible because we renamed all of the channels to make it more clear which ones are first-party vs third-party, and for various other reasons. As such every command that selects a specific conda channel (`-c conda-forge`) didn't work for me.

I think I got local dev working regardless but it wasn't painless.

conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -704,7 +704,7 @@ def render(self) -> str:
         if self._schema_version == SchemaVersion.V0:
             for key, val in self._vars_tbl.items():
                 # Double quote strings
-                if isinstance(val, str):
+                if isinstance(val, str) and not val.startswith("env.get("):
```
schuylermartin45, line 707: Any reason why this variant captures the `(` and the other doesn't? If we don't go with a regex, should we at least make this a constant string.
```suggestion
                # See Issue #271 for details on the `env.get` case.
                if isinstance(val, str) and not val.startswith("env.get("):
```

conda_recipe_manager/parser/recipe_parser_convert.py
```diff
@@ -132,7 +132,9 @@ def _upgrade_jinja_to_context_obj(self) -> None:
                 self._msg_tbl.add_message(MessageCategory.WARNING, f"The variable `{name}` is an unsupported type.")
                 continue
             # Function calls need to preserve JINJA escaping or else they turn into unevaluated strings.
-            if isinstance(value, str) and search_any_regex(Regex.JINJA_FUNCTIONS_SET, value):
+            if isinstance(value, str) and (
+                search_any_regex(Regex.JINJA_FUNCTIONS_SET, value) or value.startswith("env.get")
+            ):
```
schuylermartin45, line 137: I'm wondering if we should use a regex over `startswith` to be more flexible. I might be over indexing on this as I _think_ `value` is stripped/sanitized at this point. But I've seen enough strange recipes developing this converter that I wouldn't be surprised if there are edge cases that break with `startswith`.

I'll add this breadcrumb comment:
```suggestion
            # See Issue #271 for details on the `env.get` case.
            if isinstance(value, str) and (
                search_any_regex(Regex.JINJA_FUNCTIONS_SET, value) or value.startswith("env.get")
            ):
```

schuylermartin45 (changes requested):
I wouldn't worry about this being "hacky". The V0 recipe format is a mess, and to some extent, this library pays the price to be inherently messy so others don't have to be :)

AaronOpfer:
You touched on it in a few places in your review so I will address it here:

The main problem that caused me to introduce all of these `startswith` hacks is that variables that are going to be moved into `context:` appear to lose information about whether their values were originally _strings_ or _expressions_; this is then compensated for using a regex that seems to catch the common jinja operators. This has unhappy implications, such as the fact that the difference between `{% set var = "env.get(...)" %}` and `{% set var = env.get(...) %}` (string vs expression) are lost. This PR adds a hacky assumption that nobody ever is interested in setting a string that starts with `env.get`.

Ideally, we would instead make the recipe's context values a  `Union[ConstantStr, Expression]` such that we don't have to create these strange special cases where we bail out of escaping, since these are inherently fragile.

I am not sure how much appetite you have for this though.

schuylermartin45:
> * The `make dev` fails on up-to-date tooling if the `conda-recipe-manager` environment doesn't already exist as the conda command exits non-zero. I had to create a token env first to satisfy the makefile.
> * The makefile doesn't work on my ancient CentOS7 `make` version 3.82. I needed to source `make` from `conda-forge` before I could proceed.

If you could give more details on which version(s) of `make` were having issues, I would appreciate it. I think that `bmake` (Berkeley variant) has had some compatibility issues we've seen in the past. We install a known working `make` to environment to try to avoid some of this, but there's obviously a bootstrapping issue on first setup. At some point we will probably add `pixi` support to streamline this as an alternative setup mechanism. There is an outstanding PR that Wolf and I haven't had time to circle back on: #11 

I suppose one could argue that we should edit the `Makefile` to be compatible with all `make` variants/versions. But I've run into many headaches in the past trying to get conda environments to play nice with `make`. It is not a fun time.

> * Various tooling seems to be straddling `defaults` and `conda-forge` which is a no-no.

That's a stop-gap for waiting on packages to be made available. IMO that shouldn't have to be an issue, or at least conda should throw a fit if that is unsupported behavior. I'll have to have a think if we should prefer `conda-forge` over `defaults`.

> * Attempting to use the assisted build inside my company's conda configuration was impossible because we renamed all of the channels to make it more clear which ones are first-party vs third-party, and for various other reasons. As such every command that selects a specific conda channel (`-c conda-forge`) didn't work for me.
> 
> I think I got local dev working regardless but it wasn't painless.

That's upsetting to hear, but not surprising with being a trailblazer with this PR. I'm not super familiar with such configurations. I personally maintain the `cond-forge` release for CRM. I believe another team inside Anaconda has started packaging it for `defaults`, but it will not be updated on the same cadence.

Thanks for the feedback!

AaronOpfer:
I added test cases and I resolved your comments by adding breadcrumbs and making all references to `env.get(` use the same variation (`env.get(`). Let me know if there is anything else needed.

schuylermartin45 (approved):
As an aside: it looks like we reduced our error count in the `rattler-build` conversion integration testing by 4. Nice.

Thanks for the contribution!

PR #274, opened 2024-12-17
244 bump recipe should exit and save gracefully on some failures

schuylermartin45:
Adds a new `--save-on-failure` flag to `crm bump-recipe` to allow the tool to save partially upgraded files, in the event of an unrecoverable failure.

Also simplifies adding future flags by introducing a `_CliArgs` struct to easily pass new flags into all necessary functions going forward.

### DIFF
--- a/conda_recipe_manager/commands/bump_recipe.py
+++ b/conda_recipe_manager/commands/bump_recipe.py
@@ -9,7 +9,7 @@ import logging
 import sys
 import time
 from pathlib import Path
-from typing import Final, Optional, cast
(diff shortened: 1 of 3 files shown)


schuylermartin45:
Looks like rattler-build put out a new version last night, and that's causing some of the integration testing thresholds to fail. Will need to create a new issue to investigate, but this shouldn't be a blocker for this PR. There is no recipe conversion code being changed.

JeanChristopheMorinPerso:
I was about to as for that feature in private :)

PR #281, opened 2025-01-03
Fixes ambiguous conversion issue

### LINKED ISSUE #276: Investigate Integration test degradation with `rattler-build@v0.33.0`
As discovered in PR #274, the latest version of `rattler-build` degrades our compatibility to the point where we no longer pass some integration tests.

It is good to see that this early alert system is working as intended! So we need to go in and figure out what happened and if we need to make any improvements to maintain our current level of compatibility.

schuylermartin45:
Fixes #276 

I'm not entirely thrilled with how complex this has become.

Improvements:
- Increases `crm convert`'s compatibility with `rattler-build` by automatically upgrading most ambiguous dependencies
  - Dependencies that contain JINJA variable substitutions are skipped, but a warning is produced
- `crm convert` now will dump a full stack trace of any exception thrown when the `--debug` flag is used
- Adds `RecipeReader::is_python_recipe()` to standardize checking for the vague question "is this a Python recipe?"
  - This also fixes an existing issue with the `pip_check` upgrade path
- Fixes the GHA checks that were failing due to a new known conda telemetry issue.
- Adds some very minimal effort support for substituting JINJA variables containing ternary statements
- Adds a number of new unit and regression tests


schuylermartin45:
The known [conda telemetry issue](https://github.com/anaconda/conda-anaconda-telemetry/issues/87) is causing the the `commit check` stage to fail. See

schuylermartin45:
It seems that there's probably an edge case in `is_python_recipe()` that is now causing a few recipes to throw an exception in the conversion phase. Will investigate next.

schuylermartin45:
We're at the point of this PR where we've increased the amount of recipes that pass `crm convert`. However, the %-check used in the dry-run rattler-build phase now is not high enough. Improving the conversion success rate means we see enough new errors to cause the tests to fail.

I hope this can get resolved soon, but tracking these issues down is not always easy.

schuylermartin45:
We're at the point of diminishing returns with this PR. The remaining new issues are recipes that have dependencies with multiple operators that contain JINJA variables. There are only a handful of those, but it is enough to cause some of the bioconda tests to fail.

At this point, I want to merge this in as the PR has already ballooned to include a bunch of fixes discovered by the investigation. I have a few ideas for better improving the compatibility metrics, but I don't want to make this giant PR any worse. So I'm going to call it and merge what we have and work on some other issues.

PR #282, opened 2025-01-06
`conda-forge::` removed from environment.yaml

schuylermartin45:
`pytest-socket` is now widely available on both `conda-forge` and `defaults`, allowing us to make this change.

### DIFF
--- a/environment.yaml
+++ b/environment.yaml
@@ -10,7 +10,7 @@ dependencies:
   - python >=3.11
   - pytest
   - pytest-cov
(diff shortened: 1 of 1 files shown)


schuylermartin45:
The package on `defaults` appears to work. I added this:
```py
    import requests
    requests.get("https://google.com")
```
to one of our test cases locally. The test failed with the socket exception, as anticipated.

schuylermartin45:
NOTE: The integration testing failures are a [known issue](#276).

Issue #279, opened 2024-12-31
`convert` gives faulty `tests` section


lucascolley:
**What happened?**
This diff was needed in order to fix tests:

```diff
diff --git a/recipe/recipe.yaml b/recipe/recipe.yaml
index 548f17c..372e4c3 100644
--- a/recipe/recipe.yaml
+++ b/recipe/recipe.yaml
@@ -31,9 +31,8 @@ tests:
   - python:
       imports:
         - array_api_extra
-  - files:
-      source:
-        - tests
+  - script:
+      - pytest -ra tests
     requirements:
       run:
         - python ${{ python_min }}.*
@@ -41,9 +40,9 @@ tests:
         - pytest
         - array-api-strict
         - numpy
-    script:
-      - pip check
-      - pytest -ra tests
+    files:
+      source:
+        - tests/
 
 about:
   summary: Extra array functions built on top of the array API standard.

```

**To Reproduce**

- run `crm convert` on https://github.com/conda-forge/array-api-extra-feedstock/blob/e9fe7f3a6a0b3cbfbc33aaa167e0b020fac7e916/recipe/meta.yaml
- https://github.com/conda-forge/array-api-extra-feedstock/pull/8

**Additional Details (please complete the following information):**
version 0.4.0

schuylermartin45:
I'll have to do a little bit more digging on this. I'm very confused as to how `pip check` remained post-conversion. At first glance, it almost looks like the conversion didn't run at all.

schuylermartin45:
So in the last week or two I've made a number of compatibility improvements against the latest version of `rattler-build`. And I very recently updated the logic around some of the test section work.

If I take the original `meta.yaml` ticket provided and run it against my dev version of `crm convert`, this is the V1 `/tests` section I get:
```yaml
tests:
  - python:
      imports:
        - array_api_extra
      pip_check: true
  - files:
      source:
        - tests
    requirements:
      run:
        - python ${{ python_min }}
        - pip
        - pytest
        - array-api-strict
        - numpy
    script:
      - pytest -ra tests
```
Which looks more in alignment to the green parts of the diff listed above. A dry-run of `rattler-build` appears to confirm the issue has been resolved:
```

 ╭─ Finding outputs from recipe
 │ Found 1 variants
 │ Build variant: array-api-extra-0.4.0-pyh4616a5c_0
 │ 
 │ ╭─────────────────┬─────────╮
 │ │ Variant         ┆ Version │
[...]

PR #283, opened 2025-01-06
Redacts excessive newline

schuylermartin45:
As discussed in #279, `crm convert` no longer emits an extra blank line at the end of a rendered recipe file. This is to better conform with other automated tools in the community.


schuylermartin45:
The test failure is from an existing issue that is currently being addressed.

PR #288, opened 2025-01-15
Pre-parse Text Formatting

schuylermartin45:
The parser makes a number of assumptions areound the structure of the recipe file. This works for the vast majority of recipe files in the ecosystem, but sometimes it isn't perfect.

Ideally we enforce a standard set of format rules for our YAML files. The tricky bit is that because the V0 format is not purely YAML parseable, we can't use something like `yamlfmt`.

As a work around, I'm making a very lazy text formatter that is custom to the issues I most commonly see in our ecosystem. It isn't an ideal solution, but I hope it can better improve our parsing compatibility dramatically by making a handful of easy fixes in a pre-processing stage.

### DIFF
--- a/conda_recipe_manager/parser/_types.py
+++ b/conda_recipe_manager/parser/_types.py
@@ -128,6 +128,9 @@ class Regex:
     # Pattern to detect Jinja variable names and functions
     _JINJA_VAR_FUNCTION_PATTERN: Final[str] = r"[a-zA-Z0-9_\|\'\"\(\)\[\]\, =\.\-~\+]*"
 
+    ## V0 Formatter regular expressions ##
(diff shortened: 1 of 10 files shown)


schuylermartin45:
Good preliminary results. The first formatter change is already getting ~30 new recipes into the rattler-build stage that previously did not pass the parsing stage.

I anticipate a more dramatic change when I address the incorrectly indented test sections commonly seen in a number of R packages.

schuylermartin45:
After a few more iterations, I'm happy with the results. Just doing a handful of simple (yet arguably hacky) formatting has improved our compatibility enough to bring back one of the integration tests, and gotten the other one very close to passing.

I think fixing some of the recent user issues may get us over the line with the latest rattler changes.

PR #300, opened 2025-01-30
add `--override-build-num` flag

ForgottenProgramme:
### Description

- [ ] add `--override-build-num` flag
- [ ] ensure it cannot be used with the `--build-num` flag
- [ ] add tests

### DIFF
--- a/conda_recipe_manager/commands/bump_recipe.py
+++ b/conda_recipe_manager/commands/bump_recipe.py
@@ -49,6 +49,7 @@ class _CliArgs(NamedTuple):
     recipe_file_path: str
     # Slightly less confusing name for internal use. If we change the flag, we break users.
     increment_build_num: bool
+    override_build_num: int
     dry_run: bool
     target_version: Optional[str]
     retry_interval: float
@@ -196,6 +197,20 @@ def _update_build_num(recipe_parser: RecipeParser, cli_args: _CliArgs) -> None:
     _exit_on_failed_patch(
         recipe_parser, cast(JsonPatchType, {"op": "add", "path": _RecipePaths.BUILD_NUM, "value": 0}), cli_args
     )
+
+    if cli_args.override_build_num and recipe_parser.contains_value(_RecipePaths.BUILD_NUM):
+        build_number = recipe_parser.get_value(_RecipePaths.BUILD_NUM)
+        if not isinstance(build_number, int):
+            _exit_on_build_num_failure("Build number is not an integer.")
+
+        _exit_on_failed_patch(
+            recipe_parser,
+            cast(
+                JsonPatchType, {"op": "replace", "path": _RecipePaths.BUILD_NUM, "value": cli_args.override_build_num}
+            ),
+            cli_args,
+        )
+        return
 
 
 def _update_version(recipe_parser: RecipeParser, cli_args: _CliArgs) -> None:
@@ -376,6 +391,13 @@ def _update_sha256(recipe_parser: RecipeParser, cli_args: _CliArgs) -> None:
 @click.command(short_help="Bumps a recipe file to a new version.")
 @click.argument("recipe_file_path", type=click.Path(exists=True, path_type=str))
 @click.option(
+    "-o",
+    "--override-build-num",
+    nargs=1,
+    type=int,
+    help="Reset the build number to a custom number.",
+)
+@click.option(
     "-b",
     "--build-num",
     is_flag=True,
@@ -418,6 +440,7 @@ def _update_sha256(recipe_parser: RecipeParser, cli_args: _CliArgs) -> None:
 def bump_recipe(
     recipe_file_path: str,
     build_num: bool,
+    override_build_num: int,
     dry_run: bool,
     target_version: Optional[str],
     retry_interval: float,
@@ -433,14 +456,17 @@ def bump_recipe(
     cli_args = _CliArgs(
         recipe_file_path,
         build_num,
+        override_build_num,
         dry_run,
         target_version,
         retry_interval,
         save_on_failure,
     )
 
-    if not build_num and target_version is None:
-        log.error("The `--target-version` option must be set if `--build-num` is not specified.")
+    if not (build_num or override_build_num) and target_version is None:
+        log.error(
+            "The `--target-version` option must be set if `--build-num`, or `--override-build-num` is not specified."
+        )
         sys.exit(ExitCode.CLICK_USAGE)
 
     try:


schuylermartin45 (changes requested):

conda_recipe_manager/commands/bump_recipe.py
```diff
@@ -192,11 +193,27 @@ def _exit_on_build_num_failure(msg: str) -> NoReturn:
...
+
+    #     _exit_on_failed_patch(
+    #         recipe_parser,
+    #         cast(
+    #             JsonPatchType, {"op": "replace", "path": _RecipePaths.BUILD_NUM, "value": cli_args.override_build_num}
+    #         ),
+    #         cli_args,
+    #     )
+    #     return
+
```
schuylermartin45, line 216:
```suggestion

```
We should not get in the habit of committing commented code to `main`.

conda_recipe_manager/commands/bump_recipe.py
```diff
@@ -430,17 +456,29 @@ def bump_recipe(
     """
     # Typed, immutable, convenience data structure that contains all CLI arguments for ease of passing new options
     # to existing functions.
+    print(override_build_num)
```
schuylermartin45, line 459:
```suggestion
```

We should not be committing debug statements to `main`. One trick I've learned over the years is to mark all debug output with a `# TODO rm` or a `print(f"TODO rm {override_build_num}")` to make it easier to find all such lines before I commit/make a PR.

Ideally we'd use a debugger, but we don't live in an ideal world.

conda_recipe_manager/commands/bump_recipe.py
```diff
@@ -433,14 +456,25 @@ def bump_recipe(
...
         save_on_failure,
     )
 
-    if not build_num and target_version is None:
-        log.error("The `--target-version` option must be set if `--build-num` is not specified.")
+    if not (build_num or override_build_num) and target_version is None:
+        log.error(
+            "The `--target-version` option must be set if `--build-num`, or `--override-build-num` is not specified."
+        )
+        sys.exit(ExitCode.CLICK_USAGE)
```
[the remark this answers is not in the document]
schuylermartin45: I'm going to be a hard ass on this one (I apologize if this comes across poorly in text) and say "not sure" and "because the tests are failing" are not good answers.

We should be striving for more. If a test fails, we should ask and understand _why_ it is failing and not just leave it at "because". Leaving it 

I don't like referencing corporate slogans, but Amazon's concept/description of a [deep dive](https://www.amazon.jobs/content/en/our-workplace/leadership-principles) is the best example/write-up I can find right now about this mechanism.

tests/commands/test_bump_recipe.py
```diff
@@ -155,6 +155,91 @@ def test_bump_recipe_cli(
...
+        ("simple-recipe.yaml", "0.10.8.6", "100", "bump_recipe/build_num_100.yaml"),
+        ("simple-recipe.yaml", "0.10.8.6", "42", "bump_recipe/build_num_42.yaml"),
+        ("simple-recipe.yaml", "0.10.8.6", "0", "bump_recipe/build_num_0.yaml"),
+    ],
+)
+def test_bump_recipe_override_build_num(
+    fs: FakeFilesystem, recipe_file: str, version: str, build_num: str, expected_recipe_file: str
+) -> None:
+    """
+    Test that the `--override-build-num` flag works
```
schuylermartin45, line 170:
```suggestion
    Test that the `--override-build-num` flag allows the user to set the `/build/number` field to a positive integer.
```

tests/commands/test_bump_recipe.py
```diff
@@ -155,6 +155,91 @@ def test_bump_recipe_cli(
...
+    # Ensure that we don't check against the file that was edited.
+    assert recipe_file_path != expected_recipe_file_path
+    # Read the edited file and check it against the expected file. We don't parse the recipe file as it isn't necessary.
+    assert load_file(recipe_file_path) == load_file(expected_recipe_file_path)
+    assert result.exit_code == ExitCode.SUCCESS
+
+
+def test_bump_recipe_override_build_num_negative() -> None:
+    """
+    Ensures that negative ints are not allowed with the `--override-build-num` flag.
```
schuylermartin45, line 198:
```suggestion
    Ensures that negative integers are not allowed with the `--override-build-num` flag.
```
Nitpick

schuylermartin45 (changes requested):

schuylermartin45:
Heads up, I hit the "update PR" button so you'll need to pull to your local.

conda_recipe_manager/commands/bump_recipe.py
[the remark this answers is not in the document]
ForgottenProgramme: Yeah, sorry about that. I was a bit sloppy there. 

I went through the logic again and I think that line 468 needs to stay. What is it checking is that if `-o` and `-b` are not passed then `-t` must be passed. It's not a newly added logic check FWIW. It was already there for `-b`, I only modified it to be "either `-o` or `-b`". See [this](https://github.com/conda-incubator/conda-recipe-manager/blob/main/conda_recipe_manager/commands/bump_recipe.py#L442).

461 checks if `b` and `o` are passed together since that is not allowed.
465 checks if `o` was passed without `-t`. 

Therefore, I don't think there is duplication/repitation.

ForgottenProgramme:
I went through the logic again and I think that line 468 (in `conda_recipe_manager/commands/bump_recipe.py`) needs to stay. 

What is it checking is that if -o and -b are not passed then -t must be passed. It's not a newly added logic check FWIW. It was already there for -b, I only modified it to be "either -o or -b". See [this](https://github.com/conda-incubator/conda-recipe-manager/blob/main/conda_recipe_manager/commands/bump_recipe.py#L442).

461 checks if b and o are passed together since that is not allowed.
465 checks if o was passed without -t.

Therefore, I don't think there is duplication/repitation

schuylermartin45:
@ForgottenProgramme I think we're talking past each other and will need to setup a meeting about this. Something isn't right here about that logic.

Checking the truthiness of `override_build_num` will not work as intended now that we default `override_build_num` to 0. In other words, any `if override_build_num` statements cannot discern between the case where the user sets the flag to `--over-build-num 0` and the case where the user does not provide the option at all.

Case in point, these two examples should not produce the same error message:
```
> crm bump-recipe --override-build-num 0  tests/test_aux_files/types-toml.yaml
2025-02-19 09:05:15,956 [ERROR] [bump_recipe]: The `--target-version` option must be set if `--build-num`, or `--override-build-num` is not specified.
```

```
> crm bump-recipe tests/test_aux_files/types-toml.yaml 
2025-02-19 09:13:04,897 [ERROR] [bump_recipe]: The `--target-version` option must be set if `--build-num`, or `--override-build-num` is not specified.
```

That's very confusing wording for the user. I think this why I said originally that we don't need to change this check, because it would explicitly tell the user what the corrective action was.

schuylermartin45:
To discern whether or not the command came from the user, we could default `--override-build-num` to `None` and then set the `CLIArgs` equivalent value to `0` if `None` was passed in. To make this less confusing, we should perform the argument validation checks before we set the `cli_args` variable.

Or we could take this approach, but I think it is a little more than what we need:
https://stackoverflow.com/questions/75512527/python-click-determine-whether-argument-comes-from-default-or-from-user

I know this is being pedantic but ensuring all edge cases are covered correctly is the path to building reliable software.

schuylermartin45:
@ForgottenProgramme Added a missing check that neither of us noticed. Also added a regression test for it.

schuylermartin45 (approved):

ForgottenProgramme (commented):
Looks good!

PR #302, opened 2025-02-12
Fixed link to cep-0014.md (from cep-14.md)

bkreider:
CEPS now have leading zeros:  cep-14.md => cep-0014.md.  This fixes the link.


schuylermartin45 (approved):
Shame they can't just redirect. Thanks for catching this!

PR #332, opened 2025-05-01
328 `crm bump-recipe` needs to correct pypi endpoints

### LINKED ISSUE #328: `bump-recipe` Needs to correct PyPi endpoints
I was looking at some logs the other day and I noticed a number of `crm bump-recipe` commands were failing to complete because the URLs in the recipe file have changed.

It seems within the last year or so, many PyPi projects now use an underscore naming convention for archive file names. For example, `pytest-cov` used to be hosted at:
```
https://pypi.io/packages/source/p/pytest-cov/pytest-cov-6.0.0.tar.gz
```

And now uses:
```
https://pypi.io/packages/source/p/pytest-cov/pytest_cov-6.1.0.tar.gz
```
Note `pytest-cov-<version>.tar.gz` -> `pytest_cov-<version>.tar.gz`

I have confirmed for multi-hyphenated packages that pattern seems to be to replace all hyphens before the final one indicating a version number. This naming convention does not apply to other parts of the URL where the package name may be present.

This is turning out to be a big enough failure point for `crm bump-recipe` that we should address it. This isn't ideal, but on first failure of a PyPi archive URL, we could attempt to retry and pull the archive at a predicted location with the hyphens swapped out.

schuylermartin45:
Apologies for the large PR.

This adds logic to correct PyPi URLs that have failed to fetch by attempting to change the file name. This is done that preserves JINJA variables as to have the least amount of "blast radius" for the user.

This also adds:
-  A small wrapper library for making PyPi API calls, adjusted and improved from a proto-CRM library.
  - The requests made provide header info in compliance with PyPi's expectations. There is no API-key for the PyPi API.
  - This also includes importing some new utility functions
- Many new automated tests and test files related to the additional files added.
- Adds a function to programmatically acquire the current version of CRM
- Refactors `bump-recipe` to track all JINJA variables in a "namespace" class. This is to prevent a "magic constants" scenario.
- Minor logging improvements
- A `null_file.txt` to use in unit testing error cases (this fixes a small bug with some HTTP mocking)


conda_recipe_manager/commands/bump_recipe.py
```diff
@@ -151,7 +176,7 @@ def _exit_on_failed_fetch(recipe_parser: RecipeParser, fetcher: BaseArtifactFetc
     """
     if cli_args.save_on_failure:
         _save_or_print(recipe_parser, cli_args)
-    log.error("Failed to fetch `%s` after %s retries.", fetcher, _RETRY_LIMIT)
+    log.error("Failed to fetch `%s` after attempted retries.", fetcher)
```
msentissi, line 179: Is the number of retries considered a useless information to log ?

schuylermartin45: Each retry is counted and logged when it is run. I didn't think it was worth passing along a counter that now changes just to log something we can already track. Granted, that's a debug message

Here's a rough example of what it looks like:
```
2025-03-24 21:09:43,004[DEBUG|AnacondaRecipes/setuptools-feedstock]: 2025-03-24 21:09:41,770 [INFO] [bump_recipe]: Fetching artifact `setuptools`, attempt #1
2025-03-24 21:09:43,004[DEBUG|AnacondaRecipes/setuptools-feedstock]: 2025-03-24 21:09:41,770 [INFO] [bump_recipe]: Fetching artifact `setuptools`, attempt #2
2025-03-24 21:09:43,004[DEBUG|AnacondaRecipes/setuptools-feedstock]: 2025-03-24 21:09:41,770 [INFO] [bump_recipe]: Fetching artifact `setuptools`, attempt #3
.....
```

msentissi (approved):
LGTM

PR #348, opened 2025-05-14
Give warning or error when bumping recipe with current version 

### LINKED ISSUE #347: Give warning or error when bumping recipe with current version
**What is the idea?**

We should probably warn or error-out, if someone gives us the same version we are already on when using `bump-recipe`.

psteyer:
### Description

Resolves: https://github.com/conda/conda-recipe-manager/issues/347

Modified the tool to exit with an error when the target version matches the current version in the recipe, preventing unnecessary changes or accidental resets of the build number to 0. The change is implemented in the `bump_recipe` function where we check the versions.

### DIFF
--- a/conda_recipe_manager/commands/bump_recipe.py
+++ b/conda_recipe_manager/commands/bump_recipe.py
@@ -758,8 +758,8 @@ def bump_recipe(
     if cli_args.target_version is not None:
         if cli_args.target_version == recipe_parser.get_value(_RecipePaths.VERSION, default=None, sub_vars=True):
             log.warning(
-                "The provided target version is the same value found in the recipe file: %s", cli_args.target_version
-            )
+                "The provided target version is the same value found in the recipe file: %s", cli_args.target_version)
+            sys.exit(ExitCode.PARSE_EXCEPTION)
 
         # Version must be updated before hash to ensure the correct artifact is hashed.
         _update_version(recipe_parser, cli_args)
--- a/conda_recipe_manager/commands/bump_recipe.py
+++ b/conda_recipe_manager/commands/bump_recipe.py
@@ -747,6 +747,11 @@ def bump_recipe(
         log.error("An error occurred while parsing the recipe file contents.")
         sys.exit(ExitCode.PARSE_EXCEPTION)
 
+    if cli_args.target_version == recipe_parser.get_value(_RecipePaths.VERSION, default=None, sub_vars=True):
+        log.error(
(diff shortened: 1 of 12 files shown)


conda_recipe_manager/commands/bump_recipe.py
```diff
@@ -758,8 +758,8 @@ def bump_recipe(
     if cli_args.target_version is not None:
         if cli_args.target_version == recipe_parser.get_value(_RecipePaths.VERSION, default=None, sub_vars=True):
             log.warning(
-                "The provided target version is the same value found in the recipe file: %s", cli_args.target_version
-            )
+                "The provided target version is the same value found in the recipe file: %s", cli_args.target_version)
+            sys.exit(ExitCode.PARSE_EXCEPTION)
```
schuylermartin45, line 762: Well this is a lot simpler than I thought. A few notes:

I would probably use `CLICK_USAGE` as the error code. It sounds a little weird, but given this is input validation, that would be the best fit (`click` is the library we use to parse and validate user input).

Looking at the logic above, I would move this check to before the `_post_process_cleanup()` so that we don't make changes on a request we deem to be in error.

I would also change this to a `log.error()` to reflect that this is critical enough to kill the program.

schuylermartin45 (changes requested):

conda_recipe_manager/commands/bump_recipe.py
schuylermartin45: And it also goes without saying that we need to make sure that al the automatic checks pass before we approve this. Let me know if you have any questions about the logs!

schuylermartin45 (commented):
I'll approve this once the commented parameter code is deleted.

tests/commands/test_bump_recipe.py
```diff
@@ -192,11 +192,53 @@ def test_bump_recipe_cli(
...
+def test_bump_recipe_cli_with_same_version(
+    fs: FakeFilesystem, recipe_file: str, version: str, user_override_flag: bool
+) -> None:
+    """
+    Test that the recipe can't be bumped to the same version.
+
+    :param fs: `pyfakefs` Fixture used to replace the file system
+    :param recipe_file: Target recipe file to update
+    :param version: Target version number
+    :param user_override_flag: provides the `--override-build-num` flag
```
schuylermartin45, line 214:
```suggestion
    :param user_override_flag: Provides the `--override-build-num` flag
```
NP/Personal preference.

schuylermartin45 (approved):

PR #349, opened 2025-05-15
Fixes JINJA rendering issue in V0 recipes

schuylermartin45:
- This fixes an issue with `RecipeReader::get_value(..., sub_vars=True)` where V0 recipes would use the V1 escape syntax to get around some type evaluation issues.
- Now unrendered V0 variables will be properly returned as unescaped JINJA statements.

### DIFF
--- a/.github/workflows/commit_checks.yaml
+++ b/.github/workflows/commit_checks.yaml
@@ -47,8 +47,9 @@ jobs:
             source $CONDA/bin/activate
             conda create --name build-env -y python=3.11 conda-build
             conda activate build-env
-            conda build -c defaults -c conda-forge recipe/meta.yaml
+            conda build -c defaults recipe/meta.yaml
   # Eat our own dog food and build this project with rattler-build by converting our existing recipe.
+  # Remember, `rattler-build` will pull from the `conda-forge` channel, which may cause some version inconsistencies.
   build-recipe-rattler:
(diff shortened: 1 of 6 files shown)


msentissi:
Are the commit checks failures expected ?

schuylermartin45:
> Are the commit checks failures expected ?

No but I'm investigating. Might be a change with `rattler-build`

schuylermartin45:
The good news is the integration tests are identical with the previous iteration. So this appears to have no effect on the conversion process.

schuylermartin45:
Ok, here's our issue, I think there's a breaking change in `click`'s newest version, 8.2.0. This runner appears to have pulled from `conda-forge` for whatever reason. Maybe that's how the `miniconda3` action is setup(?)

schuylermartin45:
The breaking change is highlighted in [their change log](https://github.com/pallets/click/blob/main/CHANGES.rst#version-820).

schuylermartin45:
Ok I think I figured it out, `rattler-build` I think only installs using `conda-forge` and this test converts the recipe file and then uses rattler-build to do a test build, we get the latest version of the recipe file. 

So unfortunately the best course of action right now is to limit click to per-8.2.0 and put in a comment indicating why this was done.

msentissi (approved):

PR #356, opened 2025-05-28
fix clean-env

msentissi:
### Description

- Starting from version 24.9.0, [conda errors out](https://github.com/conda/conda/releases/tag/24.9.0), instead of doing nothing, when attempting to remove a non-existent environment.
- This causes `clean-env` to fail the first time that `make dev` is run, since the environment doesn't exist and we attempt to remove it.

### Fix
- Check whether the environment exists. If so, remove it, otherwise move on.


Makefile
```diff
@@ -52,7 +52,9 @@ clean-build: ## Removes build and Python artifacts
 	find . -name '__pycache__' -exec rm -fr {} +
 
 clean-env:					## remove conda environment
-	conda remove -y -n $(CONDA_ENV_NAME) --all
+	if conda env list | grep -q "^$(CONDA_ENV_NAME) "; then \
+		conda remove -y -n $(CONDA_ENV_NAME) --all; \
```
schuylermartin45, line 56:
```suggestion
        # In `conda@v25.5.0`, deletion of a non-existent conda environment returns an error code.
	if conda env list | grep -q "^$(CONDA_ENV_NAME) "; then \
		conda remove -y -n $(CONDA_ENV_NAME) --all; \
```

schuylermartin45 (approved):
@msentissi Approved, contingent on you taking in my comment suggestion so we remember why we did this.

schuylermartin45 (approved):

Issue #369, opened 2025-06-10
Fix development environments


schuylermartin45:
It seems that in the last few days, `make dev` no longer creates a stable environment to work in. @psteyer and I uncovered this looking at some weird behavior locally and in our GitHub Actions.

A lot of the CRM tests now take 9+ minutes to complete and fail in areas that make no sense. The logs mention some kind of `weak_ref.py` exceptions though it is too early to tell if that indicates this issue (this might be some `pyfakefs` warning I vaguely remember seeing in the past).

I `diff`'d the packages from the `build-conda-recipe` logs between the working and bad installs. Here is that diff:

```sh
$ diff good_gha_crm_env.log bad_gha_crm_env.log 
13c13
<     conda-25.5.0               |  py311h06a4308_0         1.2 MB
---
>     conda-25.5.1               |  py311h06a4308_0         1.2 MB
28a29
>     libxcb-1.17.0              |       h9b100fa_0         430 KB
40a42
>     pthread-stubs-0.3          |       h0ce48e5_1           5 KB
45c47
<     python-3.11.11             |       he870216_0        32.9 MB
---
>     python-3.11.13             |       h1a3bd86_0        32.6 MB
55a58
>     tk-8.6.14                  |       h993c535_1         3.4 MB
60a64,67
>     xorg-libx11-1.8.12         |       h9b100fa_1         895 KB
>     xorg-libxau-1.0.12         |       h9b100fa_0          13 KB
>     xorg-libxdmcp-1.1.5        |       h9b100fa_0          19 KB
>     xorg-xorgproto-2024.1      |       h5eee18b_1         580 KB
64c71
<                                            Total:        51.7 MB
---
>                                            Total:        56.7 MB
83c90
<   conda              pkgs/main/linux-64::conda-25.5.0-py311h06a4308_0 
---
>   conda              pkgs/main/linux-64::conda-25.5.1-py311h06a4308_0 
91a99
>   expat              pkgs/main/linux-64::expat-2.7.1-h6a678d5_0 
118a127
>   libxcb             pkgs/main/linux-64::libxcb-1.17.0-h9b100fa_0 
136a146
>   pthread-stubs      pkgs/main/linux-64::pthread-stubs-0.3-h0ce48e5_1 
142c152
<   python             pkgs/main/linux-64::python-3.11.11-he870216_0 
---
>   python             pkgs/main/linux-64::python-3.11.13-h1a3bd86_0 
159c169
[...]

PR #370, opened 2025-06-11
Partial fix for Issue #369 to stop the bleeding

schuylermartin45:
This should fix the immediate issue and unblock development.


schuylermartin45:
Well....I guess I shouldn't be surprised that the Python 3.12 test can't install a 3.11 variant. I'm going to ignore that for now if everything else is happy.

Issue #378, opened 2025-06-12
`crm bump-recipe`/`RecipeReader` strips necessary quotes.


schuylermartin45:
@msentissi and others (apologies, I don't have all the GitHub handles in front of me right now) have pointed out this example of a bad recipe bump: https://github.com/AnacondaRecipes/param-feedstock/pull/13/commits/47c07ec3a8f91292fab7efda3204206d94d07dd4

Normally our parser problems involve _adding_ too many quotes.  This shows an example of the quotes being stripped when they are necessary, causing us to emit invalid YAML/V0 recipe format.

schuylermartin45:
This is some-what related to #366

PR #379, opened 2025-06-12
 Fixes #366 Seemingly unnecessary double quoting and escaping

### LINKED ISSUE #366: Seemingly unnecessary double-quoting an escaping
**What happened?**
I ran CRM to update captum from 0.7.0 to 0.8.0. The output contained this:

```- "pytest upstream/tests -k \"not test_llm_attr_without_token\" --ignore=\"upstream/tests/attr/test_data_parallel.py\""  # [osx and x86 and py==312]```

**To Reproduce**

```crm bump-recipe -d -t 0.8.0 captum-feedstock/recipe/meta.yaml```

Filing this report after talking with @schuylermartin45

### PR DESCRIPTION
**Pull request by schuylermartin45**
Fixes a number of issues around excessive quotes in JINJA statements and other support. This should significantly improve previously seen odd behavior in `crm convert` and `crm bump-recipe`.

Also adds:
- Parser support for multiline strings that use the "double-quote-and-backslash" method
- Introduces a new `NodeVar` class to allow us to support tracking comments and selectors on JINJA variables. Most of this support is for V0, but there is some for V1.
- Improves a number of critical JINJA regular experessions
- A number of bug fixes around multiline string parsing
- A minor refactor of some older test files
- Regression tests that validate all these other improvements

From the integration tests, it looks like there is a slight decline in compatibility (by maybe a few recipe files). In my investigation, it looks like the recipe files that are now failing, for the most part, already contain invalid YAML. Overall, this should be a massive improvement over the current parsing issues.

### CODE REVIEW

**Comment by schuylermartin45** [comment]
This is why we have such lovely integration tests:
```
  "exception_histogram": {
    "EXCEPTION: An exception occurred while parsing the recipe file": 238,
    "EXCEPTION: An exception occurred while converting to the new recipe file": 15
  },
```
The change in its current state tanks the success rate for parsing `AnacondaRecipes` and `conda-forge` recipe files.

**Comment by schuylermartin45** [comment]
> Nice. The PR will fix this extra double-quoting https://github.com/AnacondaRecipes/pyside6-feedstock/pull/2/files#diff-f3725a55bf339595bf865fec73bda8ac99f283b0810c205442021f29c06eea9aR2, right?

It was supposed to, but I just looked and the regression is wrong. This must have gotten lost in the fray, I'll take another look.

**Comment by schuylermartin45** [comment]
@skupr-anaconda Now we should be good. At least for the vast majority of cases.

File: conda_recipe_manager/parser/cbc_parser.py [hunk 1]
```diff
@@ -124,15 +114,16 @@ def get_cbc_variable_value(
                 raise KeyError(f"CBC variable not found: {variable}")
             return default
 
-        cbc_entries: Final[list[_CBCEntry]] = self._cbc_vars_tbl[variable]
+        cbc_entries: Final = self._cbc_vars_tbl[variable]
 
         # Short-circuit on trivial case: one value, no selector
-        if len(cbc_entries) == 1 and cbc_entries[0].selector is None:
-            return cbc_entries[0].value
+        if len(cbc_entries) == 1 and cbc_entries[0].contains_selector():
```
msentissi, line 120: Is a `not` missing here ?

msentissi:
Is this PR meant to address issue #378 as well ?

conda_recipe_manager/parser/cbc_parser.py
schuylermartin45: I think so, good catch!

msentissi (approved):

Issue #364, opened 2025-06-04
Upstream package changes capitalization of the tarball between versions


AlbertDeFusco:
**What happened?**
Unable to bump Cerberus from version 1.3.4 to version 1.3.7

**To Reproduce**

1. Clone at https://github.com/AnacondaRecipes/cerberus-feedstock commit `961edaa59717df619127665444cfa5fcede6706c`
2. run `crm bump-recipe -d -t 1.3.7 ./recipe/meta.yaml`

```text
2025-06-04 11:57:19,884[INFO][bump_recipe]: Fetching artifact `cerberus`, attempt #1
2025-06-04 11:57:50,218[INFO][bump_recipe]: Fetching artifact `cerberus`, attempt #2
2025-06-04 11:58:50,393[INFO][bump_recipe]: Fetching artifact `cerberus`, attempt #3
2025-06-04 12:00:20,666[INFO][bump_recipe]: Fetching artifact `cerberus`, attempt #4
2025-06-04 12:02:21,074[INFO][bump_recipe]: Fetching artifact `cerberus`, attempt #5
2025-06-04 12:04:51,483[ERROR][bump_recipe]: Failed to update the SHA-256 variable.
Traceback (most recent call last):
  File "/Users/adefusco/Development/anaconda/pi-cursor/env/lib/python3.11/site-packages/conda_recipe_manager/commands/bump_recipe.py", line 472, in _update_sha256_check_hash_var
    sha, url = _get_sha256_and_corrected_url(recipe_parser, _RecipePaths.SINGLE_URL, src_fetcher, cli_args)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/adefusco/Development/anaconda/pi-cursor/env/lib/python3.11/site-packages/conda_recipe_manager/commands/bump_recipe.py", line 427, in _get_sha256_and_corrected_url
    _fetch_archive(fetcher, cli_args)
  File "/Users/adefusco/Development/anaconda/pi-cursor/env/lib/python3.11/site-packages/conda_recipe_manager/commands/bump_recipe.py", line 336, in _fetch_archive
    raise FetchError(f"Failed to fetch `{fetcher}` after {retries} retries.")
[...]

PR #382, opened 2025-06-16
Confirming case distinction

psteyer:
### Description

Addressing Issue #364 

- updating regex to handle more edge cases 
- adding test


schuylermartin45 (changes requested):

tests/commands/test_bump_recipe.py
```diff
@@ -513,3 +513,28 @@ def test_new_line_removal(
...
+def test_pypi_url_regex_matching(url: str, should_match: bool) -> None:
+    """
+    Test that the PYPI_URL regex pattern correctly identifies PyPI source URLs.
+
+    :param url: URL to test against the PyPI regex pattern
+    :param should_match: Whether the URL should match the PyPI pattern
+    """
+    # pylint: disable=protected-access
+    pypi_match = bump_recipe._Regex.PYPI_URL.match(url)
+    assert (pypi_match is not None) == should_match
```
schuylermartin45, line 540: Generally speaking, we shouldn't be testing private members. Private members are expected to change. It is the public API that we want to ensure that works.

In the future, for all we know, we may implement this URL check some other way. Then we will cause more work for ourselves because we used a bunch of non-public features.

I think we should write this in the style of the other tests that uses mocked URLs and mocked test files. Arguably it isn't really a unit test at that point, and more just a test of the CLI, but it should be more stable and follows more "best practices".

schuylermartin45: I believe this comment still applies. I think we should get rid of this test in favor of the other tests.

conda_recipe_manager/commands/bump_recipe.py
```diff
@@ -379,36 +378,17 @@ def _correct_pypi_url(recipe_reader: RecipeReader, url_path: str) -> tuple[str,
     #     changes, less disruption/confusion for our package builders (or so the theory goes).
     #  2) The PyPi API generally produces a "post-redirect" URL for package artifacts. Again, as to not change the file
     #     dramatically, we want to keep the commonly used `https://pypi.io` URL.
```
schuylermartin45, line 380: I can't select the whole comment in the diff, but this comment needs to be updated with the reasoning behind the new logic. We should also link-back to the original ticket for context.

conda_recipe_manager/commands/bump_recipe.py
```diff
@@ -379,36 +378,17 @@ def _correct_pypi_url(recipe_reader: RecipeReader, url_path: str) -> tuple[str,
...
-            ),
-        )
-        return (fetcher_url, f"{base_original_url}/{filename_with_var}")
+    filename: Final = pypi_meta.releases[version_value].filename
+    name: Final = pypi_meta.info.name
+    new_url: Final = f"https://pypi.org/packages/source/{ name[0] }/{ name }/{ filename }"
 
-    return (fetcher_url, f"{base_original_url}/{filename}")
+    if new_url is None:
+        raise FetchError("Unable to determine download url for PyPi API request.")
```
schuylermartin45, line 386: This check needs to happen sooner has to check if `name` and `filename` are both _non-empty_. `new_url` uses both of these variables and will produce a pretty useless URL if either or both variable is empty.

Personally I would do something like:
```
if not (filename and name):
  raise FetchError

new_url = ...
return return_url
```

Also accessing `""[0]` will cause an index exception.

schuylermartin45 (commented):
A few small notes. Also did we ever try running the `crm bump-recipe` to recreate the original issue locally? It might not be a bad idea to ensure it works in the real world at least once against that `Cerberus` package.

conda_recipe_manager/commands/bump_recipe.py
psteyer: Added a comment, Hopefully it makes enough sense. Added issue ticket link also.

conda_recipe_manager/commands/bump_recipe.py
```diff
@@ -374,41 +373,21 @@ def _correct_pypi_url(recipe_reader: RecipeReader, url_path: str) -> tuple[str,
...
-            version_value,
-            (
-                # NOTE: The double-escaping of the outer `{{` and `}}` braces.
-                f"{{{{ {_RecipeVars.VERSION} }}}}"
-                if recipe_reader.get_schema_version() == SchemaVersion.V0
-                else f"${{{{ {_RecipeVars.VERSION} }}}}"
-            ),
-        )
-        return (fetcher_url, f"{base_original_url}/{filename_with_var}")
+    # Using the PyPI metadata we can construct a url that will work for the fetcher. Incase the recipe name is
```
schuylermartin45, line 376:
```suggestion
    # Using the PyPI metadata we can construct a url that will work for the fetcher. In case the recipe name is
```
Typo

schuylermartin45 (approved):
I committed a few little comment changes, but I think this looks good now. I'll merge when this passes.

Issue #380, opened 2025-06-13
Support `-h` for help menu in all `click` commands


schuylermartin45:
This has bothered me for a long time, and I finally found the documentation on how to do this: https://click.palletsprojects.com/en/stable/documentation/#help-parameter-customization

So we should support `crm -h/--help` and `crm <command> -h/--help` for all commands

psteyer:
Completed in PR #391

PR #391, opened 2025-06-27
Adding click `help` support to click commands

psteyer:
### Description

Resolving issue #380 
- adding CONTEXT_SETTINGS to the commands to allow both `-h`, and `--help` for `crm` and `crm <command>`


msentissi:
Since the `CONTEXT_SETTINGS` is the same, can we define it in one place and import it everywhere we need it ?

msentissi (approved):

schuylermartin45 (commented):
This should be marked as `Final`

PR #392, opened 2025-06-30
assume wrong pypi url after 1 request failure

msentissi:
### Description

- There seems to be an extra sleep that isn't required after the last retry in `_fetch_archive`, so we add a check to skip it.
- When updating a recipe with an obsolete `pypi` URL using `crm bump-recipe`, the command will attempt to reach the obsolete URL twice before moving on to trying a correction, we decrease the retry interval from `30s` to `10s` to improve user experience while hopefully maintaining a long enough backoff interval.

### DIFF
--- a/conda_recipe_manager/commands/bump_recipe.py
+++ b/conda_recipe_manager/commands/bump_recipe.py
@@ -335,7 +335,8 @@ def _fetch_archive(fetcher: HttpArtifactFetcher, cli_args: _CliArgs, retries: in
             fetcher.fetch()
             return
         except FetchError:
-            time.sleep(retry_id * cli_args.retry_interval)
+            if retry_id < retries:
+                time.sleep(retry_id * cli_args.retry_interval)
 
     raise FetchError(f"Failed to fetch `{fetcher}` after {retries} retries.")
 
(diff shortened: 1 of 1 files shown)


msentissi:
The `pyfakefs` setup error is tracked [here](https://github.com/pytest-dev/pyfakefs/issues/1191).

JeanChristopheMorinPerso:
>I initially considered transmitting the error code as well when the exception is raised, but after checking that 404 is no guarantee that the file is actually missing (even though it's likely as you point out), I gave up on that solution.

If you can't trust that when a service returns a 404 the resource really doesn't exists, then you cant't trust nothing... Which doesn't make sense. That's the point I'm trying to make. you should trust the status code, unless you really have reasons not to trust them.

msentissi:
> > > > Would reducing the default `retry_interval` a bit be possible ? If so we could keep 2 tries _and_ have a relatively fast runtime.
> > > 
> > > 
> > > 30 seconds is the happy middle-ground that has worked well for me in my IoT days, but I'm not married to it. I'm sure many folks use a much smaller default. A quick Google gives lots of opinions, but I'm not finding any great answers with logic/experience behind them.
> > 
> > 
> > I see a wide array of values online as well, but most of them seem to be under `30s`, does trying `10s` seem reasonable to you ?
> 
> That sounds like something we can try. I'd abandon this PR, drop the default to 10, document/reference this discussion in a comment and see how that goes. We can always change it later.

This PR still has a useful fix, i.e. not waiting after the final retry fails.

schuylermartin45 (approved):
You should update the branch before we run the tests, but this looks good.

PR #393, opened 2025-06-30
remove license_url

msentissi:
### Description

- `license_url` is not supported by `rattler-build`, we suggest removing it to fix the `rattler-build` PR check.

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -57,7 +57,6 @@ about:
   home: https://github.com/conda/conda-recipe-manager
   license: BSD-3-Clause
   license_file: LICENSE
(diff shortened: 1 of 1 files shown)


wolfv:
great! We just removed it, indeed!.

psteyer (approved):

schuylermartin45:
Good catch while I was out! We automatically pull in the latest `rattler-build` from `conda-forge`. This wouldn't be the first time we've seen a change break a test.

I've been doing this since the start to act as an "early warning system". Sometimes we find bugs early, sometimes we file issues to `rattler-build`. This one was pretty harmless though.

Issue #399, opened 2025-07-11
Render recipe variants using a CBC file


msentissi:
**Description**
In certain cases, it can be useful to render all of the variants of a `conda` recipe from a `conda_build_config.yaml` file, replacing the JINJA variables with the appropriate values.

The request is to add functionality to perform this variant rendering to the parser.

schuylermartin45:
I've started the integration of the CBC variables into the `RecipeReader` class on this branch: https://github.com/conda/conda-recipe-manager/tree/smartin_cbc_vars_in_recipe_parser

PR #400, opened 2025-07-24
Integrates `CbcParser` class into the base recipe parsing class

schuylermartin45:
This is the first set of changes that allows the The `Recipe*` classes (with the exception of `RecipeParserConvert`) to make variable substitutions with values found in a `CbcParser` instance (from a `conda_build_config.yaml` file).

This at least partially address #399 although I think there are a number of features we need to add to the `CbcParser` class to say that we have a full integration.

I'm leaving this in draft until I finish the initial set of tests.

### DIFF
--- a/tests/parser/test_recipe_reader.py
+++ b/tests/parser/test_recipe_reader.py
@@ -305,8 +305,166 @@ def test_render_to_object(file: str, substitute: bool, expected: JsonType) -> No
     :param substitute: True to run the function with JINJA substitutions on, False for off
     :param expected: Expected value to return
     """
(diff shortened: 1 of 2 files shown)


conda_recipe_manager/parser/cbc_parser.py
```diff
@@ -122,7 +127,8 @@ def get_cbc_variable_value(
 
         for entry in cbc_entries:
             selector = entry.get_selector()
-            if selector is None or selector.does_selector_apply(query):
+            # TODO: What to do if multiple selectors apply???
```
msentissi, line 130: Could you please clarify this question with an example ?

PR #414, opened 2025-08-13
Distinguish between an empty collection and a leaf node in the parse tree

msentissi:
### Description

When parsing a file such as this, with an empty key `build:` :
```
package:
  name: test-package
  version: 1.0.0

requirements:
  build:
```
CRM will treat `build:` as a simple leaf node `build`, and render it as:
```
package:
  name: test-package
  version: 1.0.0

requirements: build
```

Since this pattern can occur and is correct, we suggest:
- Adding `Node.is_strong_leaf()`: a leaf can either be an empty key or a strong leaf.
- Using `is_strong_leaf()` in `is_single_key()`, which fixes the rendering issue.
- Using the strong_leaf/empty_key distinction wherever is_leaf() or is_empty_key() was used before. Not that's I'm not quite sure if my interventions here are sensible, especially the one in the V1 conversion code.
- We also fix failing tests as a consequence of these changes.


conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -1294,7 +1298,11 @@ def _scan_requirements(path_prefix: str = "") -> None:
                 section_path = f"{path_prefix}/requirements/{section}"
                 # Relying on `get_value()` ensures that we will only examine literal values and ignore comments
                 # in-between dependencies.
-                dependencies = cast(list[str], self.get_value(section_path, []))
+                value = self.get_value(section_path, [])
+                # Handle empty keys
+                if value is None:
+                    continue
+                dependencies = cast(list[str], value)
```
schuylermartin45, line 1305: Good catch! That is a very subtle distinction between `None` and `default`

schuylermartin45 (commented):
Changes like this are a little scary, but I think you've done a good job looking at the problem. But all of our testing safe-guards on major parsing changes indicate this is a net positive improvement.

I suspect we may have to evaluate this logic at a future date. That tends to happen with major paradigm shifts like this. But that's okay, we'll deal with it when we cross that bridge.

You'll see the comments, but I think we should revert that V1 upgrade change back to `.is_leaf()` and potentially add a MessageTable warning indicating why we skipped a selector alteration.

schuylermartin45 (approved):
There is a 1 warning difference in the conversion phase. It is small, but I think we made the right call.

PR #415, opened 2025-08-15
Fix pre-formatting

msentissi:
### Description
Two cases of recipe syntax that can be encountered motivate this fix.

The following prompts the change in `_fix_excessive_indentation`:
```
- element:
    child
```

The following prompts the change in `fmt_text()`:
```
outputs:
  - name: output1
    test:
  - name: output2
    test:
```


schuylermartin45 (commented):
I made 1 comment, but I also would like to know: shouldn't we add a regression test to prevent this issue from coming back in the future?

schuylermartin45 (approved):

PR #432, opened 2025-09-11
Support converting script_env with equal sign in value in v0

isuruf:


schuylermartin45 (commented):
We appreciate the PR! But re-reading the V1 spec, I'm not sure if this will produce valid output that `rattler-build` will accept.

I would also add to the PR description that this change is specifically about the CRM V1 conversion tooling; so folks don't get confused and think this is a change to the parser.

schuylermartin45 (approved):

PR #434, opened 2025-09-12
Add an include_test_dependencies optional arg to RecipeReaderDeps.get_all_dependencies()

### LINKED ISSUE #433: RecipeReaderDeps' get_all_dependencies() should return test dependencies
**Description**
`RecipeReaderDeps.get_all_dependencies()` doesn't return test dependencies. These are useful in many contexts and the name of the function suggests that they should be included (possibly with an opt-out).

msentissi:
### Description

This PR addresses #433, adding the ability to optionally request all dependencies including test dependencies from `get_all_dependencies()`.

Added functionality and testing for the single-output and multi-output recipe cases.


schuylermartin45 (approved):
Approved, will merge pending successful code checks

PR #440, opened 2025-09-25
Support multi-line JINJA2 and guard against unsupported statements

msentissi:
**Description**
This PR introduces multi-line JINJA2 `set` statement support to CRM. Furthermore, parsing now explicitly confirms that any encountered JINJA2 statement is supported and raises a `ParsingJinjaException` if not.
Testing is also added.

**Details**
*Current Strategy:*
1. Remove all JINJA2 (works on single-line only)
2. Parse assuming no JINJA2
3. Extract all JINJA2 variables (works on single-line only)

*Proposed Strategy:*
1. Check that all JINJA2 is supported, otherwise `raise`. The exception should contain as specific information as possible.
      - A = JINJA_V0_LINE -> JINJA_V0_MULTI_LINE
      - B = JINJA_V0_SET_LINE -> JINJA_V0_SET_MULTI_LINE
      - For each match in A, check that it is in set B, otherwise `raise`.
2. Remove all JINJA2 (incl. multi-line)
3. Parse assuming no JINJA2
4. Extract all JINJA2 variables (incl. multi-line) and support multi-line rendering.

**Testing**
- Test cases with various unsupported statements (for/if)s -> Check for exception
- Test cases with multi-line and single-line JINJA


conda_recipe_manager/parser/_types.py
```diff
@@ -128,10 +128,8 @@ class Regex:
 
     # Pattern to detect Jinja variable names and functions
     _JINJA_VAR_FUNCTION_PATTERN: Final[str] = r"[a-zA-Z0-9_\|\'\"\(\)\[\]\,\\\/ =\.\-~\+:]+"
-    # Pattern to detect optional comments or trailing whitespace. NOTE: The comment is marked as an optional matching
-    # group. Failure to mark this may cause `findall()` to return empty strings if no other group is present in the
-    # regex.
-    _JINJA_OPTIONAL_EOL_COMMENT: Final[str] = r"[ \t]*(?:#[^\n]*)?$"
+    # Pattern to detect optional comments or trailing whitespace.
+    _JINJA_OPTIONAL_EOL_COMMENT: Final[str] = r"[ \t]*(?P<comment>#[^\n]*)?$"
```
schuylermartin45, line 132: TIL you can name capture groups. Neat!

conda_recipe_manager/parser/_types.py
```diff
@@ -208,12 +206,15 @@ class Regex:
...
     JINJA_V0_SUB: Final[re.Pattern[str]] = re.compile(r"{{\s*" + _JINJA_VAR_FUNCTION_PATTERN + r"\s*}}")
-    JINJA_V0_LINE: Final[re.Pattern[str]] = re.compile(
-        r"^[ \t]*({%.+%}|{#.+#})" + _JINJA_OPTIONAL_EOL_COMMENT, flags=re.MULTILINE
+    JINJA_V0_MULTI_LINE: Final[re.Pattern[str]] = re.compile(
+        r"^[ \t]*(?P<jinja>{%.+?%}|{#.+?#})" + _JINJA_OPTIONAL_EOL_COMMENT, flags=re.MULTILINE | re.DOTALL
     )
-    JINJA_V0_SET_LINE: Final[re.Pattern[str]] = re.compile(
-        r"^[ \t]*{%[ \t]*set[ \t]*" + _JINJA_VAR_FUNCTION_PATTERN + r"[ \t]*=.+%}" + _JINJA_OPTIONAL_EOL_COMMENT,
-        flags=re.MULTILINE,
+    JINJA_V0_SET_MULTI_LINE: Final[re.Pattern[str]] = re.compile(
```
schuylermartin45, line 212: It is terrifying to try to review these. I assume you tested this with a tool of some kind (like regex101)? I guess the old tests pass, so I shouldn't be worried.

conda_recipe_manager/parser/_node_var.py
```diff
@@ -87,6 +88,12 @@ def render_v0_value(self) -> str:
             and not search_any_regex(Regex.JINJA_FUNCTIONS_SET, self._value)
         ):
             return f"'{self._value}'" if '"' in self._value else f'"{self._value}"'
+        # Render lists as multiline strings for better readability.
+        if isinstance(self._value, list):
+            repr_list = [json.dumps(item) for item in self._value]
+            nl_white_space = "\n" + " " * 4
+            comma_nl_white_space = "," + nl_white_space
```
schuylermartin45, line 95: These should all be `Final`. Also shouldn't the `4` be replaced with the `TAB_SPACE_COUNT` constant?

conda_recipe_manager/parser/exceptions.py
```diff
@@ -66,3 +53,44 @@ def __init__(self, patch: JsonPatchType):
...
+
+        :param jinja_statement: The JINJA statement that was encountered.
+        """
+        super().__init__(
+            "The recipe parser was unable to interpret the provided Conda "
+            f"recipe because of an unsupported JINJA statement: {jinja_statement.strip()}.\n"
+            "Please consider reformating the recipe file to use the supported JINJA syntax:\n"
+            "    - If using {% if %} statements, please consider replacing them with selectors.\n"
+            "    - If using {% for %} statements, especially in testing logic, "
+            "please consider using a test script instead.\n"
```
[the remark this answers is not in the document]
msentissi: It's very readable, I don't think it should be moved out of CRM since it's a CRM parsing failure that any CRM user would need to understand the causes of to know how to react. So I think it makes sense to keep this explanation here.

tests/parser/test_recipe_reader.py
```diff
@@ -1588,3 +1589,43 @@ def test_calc_sha256(file: str, expected: str) -> None:
...
+
+    if exception:
+        expected_message: Final[str] = (
+            "The recipe parser was unable to interpret the provided Conda "
+            f"recipe because of an unsupported JINJA statement: {jinja_statement}.\n"
+            "Please consider reformating the recipe file to use the supported JINJA syntax:\n"
+            "    - If using {% if %} statements, please consider replacing them with selectors.\n"
+            "    - If using {% for %} statements, especially in testing logic, "
+            "please consider using a test script instead.\n"
+        )
```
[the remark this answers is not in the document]
msentissi: I think the message is important, especially given the fact that it is customized, provides important guidance to the user, and displays the exact JINJA statement that caused an issue: `{jinja_statement}`. If we update the exception message, we'll immediately see a test failure and update the test, which is straightforward. This is not that different from most changes in tested behaviour where a change in behaviour breaks a test and prompts either a test update or reconsideration.

conda_recipe_manager/parser/_types.py
msentissi: I also just learned it, super useful!

conda_recipe_manager/parser/_node_var.py
msentissi: Definitely, fixed and fixed!

tests/parser/test_recipe_reader.py
```diff
@@ -1588,3 +1589,43 @@ def test_calc_sha256(file: str, expected: str) -> None:
...
+        with pytest.raises(ParsingException) as exc_info:
```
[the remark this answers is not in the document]
msentissi: I re-worked that exception handling a bit.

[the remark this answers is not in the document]
schuylermartin45: That double try-catch did a few things:
- Ensured that the exception got logged by the parser module
- Took any generic exception and turned it into a ParsingException. At the time this was written, we just threw random/uncontrolled exceptions and there was only 1 Parsing Exception Type. Let me take a look at the proposed change and I'll give thoughts.

conda_recipe_manager/parser/_types.py
schuylermartin45: ^ Maybe we should add these notes to a comment above these regular expressions. I think that is a good way to explain it for our future selves.

schuylermartin45:
> For `get_variable()`, I added a test case to `test_get_variable` to check for lists, it seems to handle it correctly. `get_value()` already deals with all types of valid YAML including lists, correct ? The other behaviour downstream of this that I found is V1 conversion, so I adapted that as well to include lists and added testing.

Maybe my memory is bad, but I vaguely remember skipping over support for vector/non-scalar data types because V1 did not support it at the time.

If you look at the tests around those two functions, they tend to be very complex and comprehensive. I think we really need to cover our bases here and add checks around lists of various sizes and empty lists. We should also probably check lists containing different types (like a test for a list of strings, a list of bools, a list of ints, etc)

We can't have a robust parser if we don't try all the possible combinations in our automated tests. I would take a look at the `sub_vars.yaml` tests for some inspiration. Those are very comprehensive and include some very difficult parsing situations.

tests/test_aux_files/jinja2_statements/v1_format/v1_furl.yaml
```diff
@@ -0,0 +1,66 @@
...
+# test_odd_urls: mismatch due to urlsplit() behavior with slashes.
+schema_version: 1
+
+context:
+  name: furl
+  version: 2.1.4
+  skip_broken_tests:
+    - test_hosts
+    - test_netloc
+    - test_odd_urls
```
schuylermartin45, line 11: Another thing to test: YAML supports multiple ways to define a list. We should make sure we can handle the single-line equivalent.

conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -765,14 +776,15 @@ def __init__(self, content: str):  # pylint: disable=super-init-not-called
...
-            log.exception(e1)
+            self._private_init(content=content, internal_call=False)
+        # If the expected exception is thrown, re-raise it.
+        except ParsingException:
             raise
+        # If an unexpected exception is thrown, log the full chain of exceptions,
+        # then re-raise the expected exception from it.
+        except Exception as e0:  # pylint: disable=broad-exception-caught
+            log.exception(e0)
+            raise ParsingException() from e0
```
schuylermartin45, line 787: The only nuance is that with this method, the logged exception won't have the `ParsingException` typing listed.

conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -765,14 +776,15 @@ def __init__(self, content: str):  # pylint: disable=super-init-not-called
...
-            try:
-                self._private_init(content=content, internal_call=False)
-            except Exception as e0:  # pylint: disable=broad-exception-caught
-                raise ParsingException() from e0
-        except ParsingException as e1:
-            # Log the full chain of exceptions, then re-raise the expected exception.
-            log.exception(e1)
+            self._private_init(content=content, internal_call=False)
+        # If the expected exception is thrown, re-raise it.
+        except ParsingException:
```
schuylermartin45, line 781: This should probably log too.

conda_recipe_manager/parser/recipe_parser_convert.py
```diff
@@ -149,7 +149,7 @@ def _upgrade_jinja_to_context_obj(self) -> None:
         """
         # Convert the JINJA variable table to a `context` section. Empty tables still add the `context` section for
         # future developers' convenience.
-        context_obj: dict[str, Primitives] = {}
+        context_obj: dict[str, PrimitivesNoNone | list[PrimitivesNoNone]] = {}
```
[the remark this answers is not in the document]
schuylermartin45: Or is this specifically a limitation of V1?

conda_recipe_manager/parser/_types.py
msentissi: Done! :)

tests/parser/test_recipe_reader.py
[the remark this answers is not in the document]
msentissi: I removed the `expected_message` from the test and adapted the `assert`.

conda_recipe_manager/parser/recipe_reader.py
msentissi: Fixed!

conda_recipe_manager/parser/recipe_reader.py
msentissi: Done!

msentissi:
Testing has been added and updated for `get_variable(sub_vars)`, and the implementation of the function has been simplified. `set_variable` testing has been added as well.
Note that there is a large number of failures in integration tests caused by the previously silent, now loud failures when encountering unsupported JINJA2.

tests/parser/test_recipe_reader.py
```diff
@@ -884,36 +887,42 @@ def test_contains_value(file: str, path: str, expected: bool) -> None:
         # Add/concat cases
         ("v1_format/v1_sub_vars.yaml", "/requirements/fake_run_constrained/0", True, 43),
         ("v1_format/v1_sub_vars.yaml", "/requirements/fake_run_constrained/1", True, 43.3),
-        ("v1_format/v1_sub_vars.yaml", "/requirements/fake_run_constrained/2", True, "421"),
-        ("v1_format/v1_sub_vars.yaml", "/requirements/fake_run_constrained/3", True, "421.3"),
+        ("v1_format/v1_sub_vars.yaml", "/requirements/fake_run_constrained/2", True, '${{ number + "1" }}'),
+        ("v1_format/v1_sub_vars.yaml", "/requirements/fake_run_constrained/3", True, "${{ number + '1.3' }}"),
```
schuylermartin45, line 891: If sub_vars is set to True, these should be performing the equivalent concatenations, right? This feels like a step backwards.

schuylermartin45: hmmmm that documentation is a little bit more vague than I would like. Maybe we should test and verify what `jinja2` does in the python interpreter?

It looks like `~` is supposed to concatenate by up-casting to string. It is hard to search for, but it does seem to happen with some frequency in recipe files:
- https://github.com/AnacondaRecipes/clang-compiler-activation-feedstock/blob/12c7b65d9ac40380e81a05dbab635d9453290b1c/recipe/meta.yaml#L65
- https://github.com/AnacondaRecipes/xorg-xproto-feedstock/blob/dd624cea5879919cb34480e9d928ff0d9272ab9b/recipe/meta.yaml#L3

So maybe we upgrade the tests to use `~` instead when/if we move to use `jinja2`?

msentissi: Done!

conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -510,50 +403,22 @@ def _render_jinja_vars(self, s: str) -> JsonType:
...
-                s = s.replace(match, value)
+            expression = match[start_idx:-2].strip()
+            context = {k: self.get_variable(k) for k in self._vars_tbl}
+            # If the expression can't be evaluated, skip it.
+            try:
+                env = Environment(undefined=StrictUndefined)  # type: ignore[misc]
+                compiled_expression = env.compile_expression(expression)
+                result = compiled_expression(**context)  # type: ignore[misc]
+            except Exception:  # pylint: disable=broad-exception-caught
+                continue
```
schuylermartin45, line 414: Should we be logging these silent failures?

conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -634,6 +506,13 @@ def _init_schema_version_and_sanitize_v0_yaml(self, internal_call: bool) -> tupl
         function.
 
         :param internal_call: Whether this is an internal call. If true, we cannot determine if the recipe is V0 or V1.
+        :param force_remove_jinja: Whether to force remove JINJA statements from the recipe file.
+            If this is set to True,
+                then JINJA statements will be removed from the recipe file without checking if they are valid.
+            If this is set to False,
+                then JINJA statements will be checked for validity
+                and a ParsingJinjaException will be raised if they are invalid.
```
schuylermartin45, line 514: I guess if you update the other comment we'll have to update this one too.

conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -660,30 +539,47 @@ def _init_schema_version_and_sanitize_v0_yaml(self, internal_call: bool) -> tupl
...
 
-    def _private_init(self, content: str, internal_call: bool) -> None:
+    def _private_init(self, content: str, internal_call: bool, force_remove_jinja: bool = False) -> None:
         """
         Private constructor for internal RecipeReader use. This constructor is called by `__init__()` and
         `_create_private_recipe_reader()`, with internal_call set to False and True, respectively.
 
         :param content: conda-build formatted recipe file, as a single text string.
         :param internal_call: Whether this is an internal call. If true, we cannot determine if the recipe is V0 or V1.
+        :param force_remove_jinja: Whether to force remove JINJA statements from the recipe file.
```
schuylermartin45, line 566: ...and here

conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -757,22 +653,33 @@ def _private_init(self, content: str, internal_call: bool) -> None:
         # This table will have to be re-built or modified when the tree is modified with `patch()`.
         self._rebuild_selectors()
 
-    def __init__(self, content: str):  # pylint: disable=super-init-not-called
+    def __init__(self, content: str, force_remove_jinja: bool = False):  # pylint: disable=super-init-not-called
         """
         Constructs a RecipeReader instance.
 
         :param content: conda-build formatted recipe file, as a single text string.
+        :param force_remove_jinja: Whether to force remove JINJA statements from the recipe file.
```
schuylermartin45, line 661: So many constructor docs! lol.

tests/parser/test_recipe_reader.py
```diff
@@ -736,14 +737,16 @@ def test_contains_value(file: str, path: str, expected: bool) -> None:
         ("sub_vars.yaml", "/requirements/fake_run_constrained/20", True, "types_toml"),
         ("sub_vars.yaml", "/requirements/fake_run_constrained/21", True, "TYPES_TOML"),
         # Complex split and join cases. Note that we do not replace if split/join would result in a non-string value.
+        # We also do not replace if the expression can't be evaluated.
         ("sub_vars.yaml", "/requirements/fake_run_constrained/22", True, "{{ name.split('-') }}"),
-        ("sub_vars.yaml", "/requirements/fake_run_constrained/23", True, "{{ '.'.join(name) }}"),
+        ("sub_vars.yaml", "/requirements/fake_run_constrained/23", True, "T.Y.P.E.S.-.t.o.m.l"),
```
schuylermartin45, line 742: Nice!

schuylermartin45 (commented):
I think I'm feeling more confident with this now. Most of my comments are notes on documentation and logging.

schuylermartin45 (approved):
woohoo

PR #449, opened 2025-10-16
Bump recipe library module

schuylermartin45:
This refactors the work found in the `crm bump-recipe` command into a `VersionBumper` class. 

This aims to:
- Allow other automated systems to have finer control over the update process.
- Provide better access control to the cached source artifacts for other automated systems. Again, previously, `crm bump-recipe` did not allow the caller to know anything about the underlying cached source(s).

This also:
- Introduces the `conda_recipe_manager.ops` module, which provides a location to store library modules offering more complex recipe editing tooling (like version bumping)
- Adds new context-managed utility functions for fetching all source information from a recipe file (in the form of the `from_recipe_fetch*()` functions

### DIFF
--- a/conda_recipe_manager/ops/version_bumper.py
+++ b/conda_recipe_manager/ops/version_bumper.py
@@ -17,7 +17,6 @@ from conda_recipe_manager.ops.exceptions import VersionBumperInvalidState, Versi
 from conda_recipe_manager.parser.recipe_parser import RecipeParser, ReplacePatchFunc
 from conda_recipe_manager.parser.recipe_parser_deps import RecipeParserDeps
 from conda_recipe_manager.parser.recipe_reader_deps import RecipeReaderDeps
-from conda_recipe_manager.parser.types import SchemaVersion
 from conda_recipe_manager.types import JsonPatchType, JsonType
 
 log: Final = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


schuylermartin45:
This draft now passes the existing `crm bump-recipe` tests. However, I would like to add more tests to validate the new library functionality.

conda_recipe_manager/ops/version_bumper.py
```diff
@@ -0,0 +1,509 @@
...
+            raise VersionBumperInvalidState("Can't bump a recipe to the same software version.")
+
+        # If the `version` variable is found, patch that. This is an artifact/pattern from Grayskull.
+        old_variable: Final = self._recipe_parser.get_variable(_RecipeVars.VERSION, None)
+        if old_variable is not None:
+            self._recipe_parser.set_variable(_RecipeVars.VERSION, target_version)
+            # NOTE: This is a linear search on a small list.
+            if _RecipePaths.VERSION not in self._recipe_parser.get_variable_references(_RecipeVars.VERSION):
+                # In the off-chance the key is missing, add it.
+                if not self._recipe_parser.contains_value(_RecipePaths.VERSION):
```
msentissi: If the version variable is not used in the `package/version` field but the `package/version` field exists, is that case handled here ?

conda_recipe_manager/fetcher/artifact_fetcher.py
```diff
@@ -109,3 +155,193 @@ def from_recipe(recipe: RecipeReader, ignore_unsupported: bool = False) -> Gener
...
+
+
+def _fetch_archive(fetcher: BaseArtifactFetcher, retry_interval: float, retries: int) -> _FutureFetch:
+    """
+    Fetches the target source archive (with retries) for future use.
+
+    :param fetcher: Artifact fetching instance to use.
+    :param retry_interval: Base quantity of time (in seconds) to wait between fetch attempts.
+    :param retries: Number of retries to attempt.
+    :raises FetchError: If an issue occurred while downloading or extracting the archive.
```
msentissi: `:returns:` ?

schuylermartin45: Good catch. We have a `pylint` plugin that is supposed to catch these things but it seems to work sporadically at best. I haven't figured out why yet.

msentissi (approved):

PR #450, opened 2025-10-21
CBCParser: Build Variants generation

msentissi:
### Description

This PR alters `CbcParser` and associated classes to enable `CbcParser` to produce a build variants dictionary through `generate_variants()`.
- `RecipeReader` is modified to store `floats` as strings, in order to preserve version data precision (`1.20` shouldn't become `1.2` when parsing a CBC file).
- `SelectorParser` and `SelectorQuery` are reworked to:
  - Handle build environment variables in the selector when evaluating.
  - Handle parentheses in the boolean expressions
- Tests are added/updated accordingly. In particular `generate_variants()` is tested against the build variant `dict`s produced by `conda-build`'s `variants.get_package_variants()` given the same CBC files (modulo the CBC special keys we ignore for now).

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "conda-recipe-manager" %}
-{% set version = "0.8.0" %}
+{% set version = "0.8.1" %}
 
 package:
   name: {{ name|lower }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,11 @@ Note: version releases in the 0.x.y range may introduce breaking changes.
 ### Fixed
 ### Security
 
+## [0.8.1]
+### Changed
+- `RecipeReader.get_value()` now uses `render_to_object()`, simplifying the logic and preserving retrieved value types better.
+- `RecipeReader.render_to_object()` was also reworked for increased accuracy, and to better support `get_value()`.
+
 ## [0.8.0]
 ### Added
 - The `conda_recipe_manager.ops.VersionBumper` class. This takes the core logic from `crm bump-recipe` and exposes
@@ -203,7 +208,8 @@ and continually changing.
 Migrates parser from [percy](https://github.com/anaconda-distribution/percy/tree/main)
 , ,
 
-[Unreleased]: https://github.com/conda/conda-recipe-manager/compare/v0.8.0...HEAD
+[Unreleased]: https://github.com/conda/conda-recipe-manager/compare/v0.8.1...HEAD
+[0.8.1]: https://github.com/conda/conda-recipe-manager/compare/v0.8.0...v0.8.1
 [0.8.0]: https://github.com/conda/conda-recipe-manager/compare/v0.7.1...v0.8.0
 [0.7.1]: https://github.com/conda/conda-recipe-manager/compare/v0.7.0...v0.7.1
 [0.7.0]: https://github.com/conda/conda-recipe-manager/compare/v0.6.4...v0.7.0
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -16,7 +16,7 @@ from typing import Final
 project = "Conda Recipe Manager"
 author = "various"
 # TODO make this automatic, update README
-release = "0.8.0"
+release = "0.8.1"
 
 # -- General configuration ---------------------------------------------------
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#general-configuration
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ exclude = ["tests"]
 
 [project]
 name = "conda_recipe_manager"
-version = "0.8.0"
+version = "0.8.1"
 authors = [
   { name="Anaconda, Inc.", email="distribution_team@anaconda.com" },
 ]
--- a/tests/parser/test_recipe_reader.py
+++ b/tests/parser/test_recipe_reader.py
@@ -952,6 +952,13 @@ def test_contains_value(file: str, path: str, expected: bool) -> None:
         # Check that lists of lists are retrieved correctly
         ("cbc_files/anaconda_cbc_01.yaml", "/zip_keys", False, [["python", "numpy"]]),
         ("list_collection.yaml", "/", False, {"zip_keys": [["python", "numpy"]]}),
+        # Regression test, get_value() returned None for about/summary
+        (
+            "parser_regressions/get_value_summary_none.yaml",
+            "/about/summary",
+            False,
+            "#1 quality TLS certs while you wait, for the discerning tester",
+        ),
     ],
 )
 def test_get_value(file: str, path: str, sub_vars: bool, expected: JsonType) -> None:
--- a/tests/test_aux_files/parser_regressions/get_value_summary_none.yaml
+++ b/tests/test_aux_files/parser_regressions/get_value_summary_none.yaml
@@ -1,0 +1,59 @@
+{% set name = "trustme" %}
+{% set version = "1.2.1" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: 6528ba2bbc7f2db41f33825c8dd13e3e3eb9d334ba0f909713c8c3139f4ae47f
+
+build:
+  number: 1
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  skip: true  # [py<39]
+
+requirements:
+  host:
+    - pip
+    - python
+    - hatchling
+  run:
+    - python
+    - cryptography >=3.1
+    - idna >=2.0
+
+test:
+  source_files:
+    - tests
+  imports:
+    - trustme
+  commands:
+    - pip check
+    - python -c "from importlib.metadata import version; assert(version('{{ name }}')=='{{ version }}')"
+    - pytest
+  requires:
+    - pip
+    - pytest
+    - pyopenssl
+    - service_identity
+
+about:
+  home: https://trustme.readthedocs.io
+  # Single quotes are required here to avoid YAML rendering the hash symbol as a comment.
+  summary: '#1 quality TLS certs while you wait, for the discerning tester'
+  license: Apache-2.0 or MIT
+  license_family: MIT
+  license_file:
+    - LICENSE.MIT
+    - LICENSE.APACHE2
+  description: |
+    trustme is a tiny Python package that does one thing: it gives you a fake certificate authority (CA)
+    that you can use to generate fake TLS certs to use in your tests.
+  dev_url: https://github.com/python-trio/trustme
+  doc_url: https://trustme.readthedocs.io
+
+extra:
+  recipe-maintainers:
+    - bollwyvl


tests/test_aux_files/cbc_files/zip_keys_multiple_lists.yaml
```diff
@@ -0,0 +1,32 @@
...
+  -           # [linux and armv7l]
+    - lzo    # [linux and armv7l]
+    - lz4   # [linux and armv7l]
+  -           # [linux and ppc64le]
+    - xz    # [linux and ppc64le]
+    - zstd   # [linux and ppc64le]
+  -           # [linux and s390x]
+    - liblzma   # [linux and s390x]
+    - libzstd   # [linux and s390x]
+  -           # [linux and x86_64]
```
schuylermartin45: Sanity check: Judging by the file name, the list of lists is intentional. Do these lines also need the selector?

msentissi: They may not strictly need it but it seems more coherent.

schuylermartin45:
> > RecipeReader is modified to store floats as strings, in order to preserve version data precision (1.20 shouldn't become 1.2 when parsing a CBC file).
> 
> We probably have to dive deeper on this. At a certain point, I think it might be the fault of the recipe file for not encapsulating a version as a string. A version should always be a string.
> 
> This exact issue is actually highlighted in the V1 spec: https://github.com/conda/ceps/blob/main/cep-0014.md#package-section

At the very least, maybe we force-upgrade `/package/version` to be a string in V0.

recipe/meta.yaml
```diff
@@ -29,6 +29,7 @@ requirements:
     # TODO Future: click 8.2.1 is available on defaults and conda-forge now.
     - click >=8.1.7,<8.2.0
     - conda
+    - conda-build
```
schuylermartin45: I'm a little worried if this is going to have weird implications in the future. But I'm assuming we can't really get around it unless we break conda-build into having a package of forward declares/header information.

conda_recipe_manager/parser/platform_types.py
```diff
@@ -35,17 +35,32 @@ class Arch(StrEnum):
...
     ARM_V7L = "armv7l"
-    PPC_64 = "ppc64"
     PPC_64_LE = "ppc64le"
 
 
 # Set of all Architecture options
 ALL_ARCHITECTURES: Final[set[Arch]] = set(Arch)
 
 
+class PlatformAlias(StrEnum):
```
schuylermartin45: I think this is a good idea and a good name for this. The naming conventions for Arch/Platform/OS in conda build are a little difficult to keep track of.

conda_recipe_manager/parser/platform_types.py
```diff
@@ -162,3 +173,27 @@ def get_platforms_by_os(os: OperatingSystem | str) -> set[Platform]:
...
+        alias_sanitized: Final[str] = alias.strip().lower()
+        if not alias_sanitized in ALL_PLATFORM_ALIASES:
+            return set()
+        alias = PlatformAlias(alias_sanitized)
+
+    match alias:
+        case PlatformAlias.LINUX_32:
+            return {Platform.LINUX_32}
+        case PlatformAlias.LINUX_64:
+            return {Platform.LINUX_64}
```
schuylermartin45: As a sanity check: Should these include other LINUX 64 and 32 bit platforms?

conda_recipe_manager/parser/exceptions.py
```diff
@@ -74,6 +74,35 @@ def __init__(self, node: Node):
...
+        """
+        Constructs a zip keys exception.
+
+        :param zip_keys: List of zip keys that were encountered.
+        :param message: String description of the issue encountered.
+        """
+        super().__init__(f"{message}: {zip_keys}")
+
+
+class SelectorSyntaxError(BaseParserException):
```
schuylermartin45: I may be wrong, but I could have sworn some selector parser exceptions already exist(?). If not, I think given the selector parser is in this `parser` module, this is a good place to put these exceptions.

Though I wonder if the selector/CBC exception should inherit from some base selector/CBC exception. And then those could inherit from the base parser exception. That may be a bit overkill, but it'll make the family tree more sensible if we have to add other Selector and CBC exceptions later(?)

conda_recipe_manager/parser/_types.py
```diff
@@ -279,3 +285,25 @@ class Regex:
...
+
+def string_constructor(loader: SafeLoader, node: ScalarNode) -> str:
+    """
+    Constructor that forces all float values to be strings.
+
+    :param loader: The loader to use
+    :param node: The node to construct
+    :returns: The constructed value
+    """
+    return cast(str, loader.construct_scalar(node))
```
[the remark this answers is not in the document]
msentissi: Agreed, fixed!

conda_recipe_manager/parser/cbc_parser.py
```diff
@@ -89,6 +104,55 @@ def __contains__(self, key: object) -> bool:
...
+        """
+        is_list_of_lists: Final[bool] = isinstance(value_list, list) and all(
+            isinstance(inner_list, list) and all(isinstance(elem, str) for elem in inner_list)
+            for inner_list in value_list
+        )
+        is_list_of_strings: Final[bool] = isinstance(value_list, list) and all(
+            isinstance(elem, str) for elem in value_list
+        )
+        if not is_list_of_lists and not is_list_of_strings:
+            return
```
schuylermartin45: Do we want to long a warning or throw here?

conda_recipe_manager/parser/cbc_parser.py
```diff
@@ -97,7 +161,7 @@ def list_cbc_variables(self) -> list[str]:
         """
         return list(self._cbc_vars_tbl.keys())
 
-    def get_cbc_variable_value(
+    def get_cbc_variable_values(
```
schuylermartin45: Is renaming this function going to cause a lot of breakages?

conda_recipe_manager/parser/cbc_parser.py
```diff
@@ -115,18 +179,179 @@ def get_cbc_variable_value(
...
+                selected_entries.append(entry.get_value())
+        if selected_entries:
+            return selected_entries
 
         # No applicable entries have been found to match any selector variant.
         if isinstance(default, SentinelType):
             raise ValueError(f"CBC variable does not have a value for the provided selector query: {variable}")
         return default
+
+    def _validate_zip_keys(self, zip_keys: list[set[str]]) -> None:
```
schuylermartin45: Should this be a static function? I don't think it uses `self`

conda_recipe_manager/parser/cbc_parser.py
```diff
@@ -115,18 +179,179 @@ def get_cbc_variable_value(
...
+                pass
+            except ZipKeysException as e:
+                raise e
+            for variable in cbc_file.list_cbc_variables():
+                try:
+                    cbc_values[variable] = cast(
+                        list[Primitives], cbc_file.get_cbc_variable_values(variable, selector_query)
+                    )
+                except ValueError:
+                    continue
```
schuylermartin45: Why do we ignore this case?

conda_recipe_manager/parser/selector_parser.py
```diff
@@ -105,54 +116,161 @@ def _v0_extract_selector(comment: Optional[str]) -> Optional[str]:
...
+                continue
+            if idx + 2 >= len(tokens) or tokens[idx + 2].is_operator():
+                raise SelectorSyntaxError(
+                    f"Did not find a second operand after the operator {operator} while reducing {op}: {tokens}"
+                )
+            operator.l_node = cur_operand
+            operator.r_node = tokens[idx + 2]
+            idx += 2
+            cur_operand = operator
+        return new_tokens
```
schuylermartin45: At first glance, this looks like the [Shunting Yard Algorithm](https://en.wikipedia.org/wiki/Shunting_yard_algorithm). I'm linking this here in the off chance you haven't read this wiki before.

I love this algorithm. It isn't THE Dijkstra algorithm, but it is up there.

schuylermartin45: And if you'd prefer, you could use the stack operators on Python lists instead of tracking the index.

conda_recipe_manager/parser/cbc_parser.py
```diff
@@ -89,6 +104,55 @@ def __contains__(self, key: object) -> bool:
...
+        :param comments_tbl: Table of comments.
+        """
+        # Re-assemble the comment components. If successful, append it to the node.
+        # TODO Improve: This is not very efficient.
+        selector_str = "" if not self.contains_selector_at_path(path) else self.get_selector_at_path(path)
+        comment_str = comments_tbl.get(path, "")
+        combined_comment = f"{selector_str} {comment_str}"
+        return NodeVar(value, f"# {combined_comment}" if combined_comment.strip() else None)
+
+    def _construct_zip_keys(self, value_list: list[JsonType], comments_tbl: dict[str, str]) -> None:
```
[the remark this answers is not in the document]
msentissi: Done.

conda_recipe_manager/parser/cbc_parser.py
msentissi: It should!

conda_recipe_manager/parser/cbc_parser.py
```diff
@@ -115,18 +179,179 @@ def get_cbc_variable_value(
...
+        Constructs the default variants CBC file.
+
+        :returns: The default variants CBC file.
+        """
+        default_variants = {}
+        for key, value in DEFAULT_VARIANTS.items():
+            if isinstance(value, Primitives):
+                default_variants[key] = [value]
+            else:
+                default_variants[key] = value
```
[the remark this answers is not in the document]
msentissi: Done!

conda_recipe_manager/parser/cbc_parser.py
msentissi: A `ValueError` is raised when no value can be retrieved given a particular selector query, meaning that key isn't defined in the CBC as far as some particular platform/build env config is concerned, which is valid.

conda_recipe_manager/parser/cbc_parser.py
schuylermartin45: I don't think anything is. I also ran an internal search and I didn't get any hits.

schuylermartin45: So we're probably safe.

conda_recipe_manager/parser/cbc_parser.py
```diff
@@ -115,18 +179,179 @@ def get_cbc_variable_value(
...
+                    cbc_values[variable] = cast(
+                        list[Primitives], cbc_file.get_cbc_variable_values(variable, selector_query)
+                    )
+                except ValueError:
+                    continue
+        # Validate that all zip keys are present in the CBC values
+        CbcParser._validate_zip_keys_against_cbc_values(zip_keys, cbc_values)
+        return cbc_values, zip_keys
+
+    @no_type_check
```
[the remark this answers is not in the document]
schuylermartin45: Send me some details on Slack and I'll see if I can help. Or maybe we can get on a call later. I have yet to be completely stumped by `mypy`. Life....uh..._finds a way_.

conda_recipe_manager/parser/cbc_parser.py
```diff
@@ -115,18 +179,179 @@ def get_cbc_variable_value(
...
+    @staticmethod
+    def generate_variants(cbc_files: list[CbcParser], selector_query: SelectorQuery) -> tuple[dict[str, JsonType]]:
```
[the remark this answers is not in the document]
msentissi: Done!

conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -664,7 +671,9 @@ def _private_init(self, content: str, internal_call: bool, force_remove_jinja: b
         # This table will have to be re-built or modified when the tree is modified with `patch()`.
         self._rebuild_selectors()
 
-    def __init__(self, content: str, force_remove_jinja: bool = False):  # pylint: disable=super-init-not-called
+    def __init__(
+        self, content: str, force_remove_jinja: bool = False, floats_as_strings: bool = False
```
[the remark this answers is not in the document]
msentissi: Done!

conda_recipe_manager/parser/cbc_parser.py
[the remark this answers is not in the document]
msentissi: Life did find a way, fixed!

conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -567,7 +567,9 @@ def _init_schema_version_and_sanitize_v0_yaml(
 
         return str(sanitized_fmt), tof_comment_cntr
 
-    def _private_init(self, content: str, internal_call: bool, force_remove_jinja: bool = False) -> None:
+    def _private_init(
+        self, content: str, internal_call: bool, force_remove_jinja: bool = False, floats_as_strings: bool = False
```
schuylermartin45: Do we want to just carry the flags all the way through for consistency?

conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -587,6 +591,7 @@ def _private_init(self, content: str, internal_call: bool, force_remove_jinja: b
         # Note: _init_content should be Final, but mypy requires Final attributes to be declared in __init__
         # See https://mypy.readthedocs.io/en/stable/final_attrs.html#syntax-variants
         self._init_content: str = content
+        self._yaml_loader: type[SafeLoader] = StringLoader if floats_as_strings else SafeLoader
```
schuylermartin45: Is this `Final`-able?

conda_recipe_manager/parser/recipe_reader.py
msentissi: The flags make the outer interface more elegant, but I'm not sure they simplify things when carried too far in.

schuylermartin45 (approved):
I'm approving this, with the caveats we have previously discussed:
- We will kick the can on implementing shunting yard
- We won't merge this until after the 0.8.1 release

PR #453, opened 2025-10-24
Exposes the path to the archive file managed by the `HttpArtifactFetcher` class

schuylermartin45:
- Adds `HttpArtifactFetcher::get_path_to_archive()` so callers can have access to the downloaded archive file directly in temp storage.
- This is not ideal, but some community tooling only works against archive files directly (as opposed to pointing tools at uncompressed directories)
- Adds unit tests

### DIFF
--- a/tests/fetcher/test_http_artifact_fetcher.py
+++ b/tests/fetcher/test_http_artifact_fetcher.py
@@ -269,7 +269,7 @@ def test_get_path_to_archive_raises_no_fetch(
     :param http_fetcher_failure: HttpArtifactFetcher test fixture.
     """
     with pytest.raises(FetchRequiredError):
-        http_fetcher_failure.get_path_to_source_code()
+        http_fetcher_failure.get_path_to_archive()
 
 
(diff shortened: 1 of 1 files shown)


tests/fetcher/test_http_artifact_fetcher.py
```diff
@@ -216,13 +222,51 @@ def test_get_path_to_source_code(http_fixture: str, expected_src: str, request:
...
+def test_get_path_to_archive_raises_no_fetch(
+    http_fetcher_failure: HttpArtifactFetcher,  # pylint: disable=unused-argument
+) -> None:
+    """
+    Ensures `get_path_to_archive()` throws if `fetch()` has not been called.
+
+    :param http_fetcher_failure: HttpArtifactFetcher test fixture.
     """
     with pytest.raises(FetchRequiredError):
         http_fetcher_failure.get_path_to_source_code()
```
leahsch: should this be `http_fetcher_failure.get_path_to_archive()`?

PR #455, opened 2025-10-28
use render_to_object() in get_value()

msentissi:
### Description
Previously, `get_value()` performed several separate, case-by-case, processing steps that were:
- Less readable.
- Converted the underlying AST into a string representation, then reparsed it: this lost type information in the retrieved JSON.

The following changes aim to rely heavily on `render_to_object()` in `get_value()`, greatly simplifying the logic within it.
`render_to_object()` was also updated to support this transition.

### DIFF
--- a/conda_recipe_manager/parser/_node.py
+++ b/conda_recipe_manager/parser/_node.py
@@ -219,6 +219,8 @@ class Node:
         :returns: True if the node contains a list. False otherwise.
         """
 
+        # We track both list members and non-list members to ensure the node tree is consistent.
+        # Either all children must be list members or all children must be non-list members.
         contains_list_member = False
         contains_non_list_member = False
         for child in self.children:


conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -928,6 +928,25 @@ def render(self, omit_trailing_newline: bool = False) -> str:
...
+        Handle multi-line strings and variable replacement.
+
+        :param node: Node from which to extract the value to preprocess
+        :param replace_variables: If set to True, this replaces all variable substitutions with their set values.
+        :raises TypeError: If the node value is a sentinel type
+        :returns: Preprocessed value
+        """
+        if isinstance(node.value, SentinelType):
+            raise TypeError(f"Node value is a sentinel type: {node.value}")
+        value = normalize_multiline_strings(node.value, node.multiline_variant)
```
schuylermartin45:
```suggestion
        value: Final = normalize_multiline_strings(node.value, node.multiline_variant)
```

conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -928,6 +928,25 @@ def render(self, omit_trailing_newline: bool = False) -> str:
 
         return "\n".join(lines)
 
+    def _preprocess_node_value(self, node: Node, replace_variables: bool) -> JsonType:
+        """
+        Handle multi-line strings and variable replacement.
+
+        :param node: Node from which to extract the value to preprocess
+        :param replace_variables: If set to True, this replaces all variable substitutions with their set values.
+        :raises TypeError: If the node value is a sentinel type
```
schuylermartin45: You mention the `raises` here in the docs, but you don't include it in the public facing function. So callers won't know by reading the auto-generate public docs that this is a possibility.

I also wonder if we should use one of our customer parsing exceptions or invent a new one for this case. Ideally, parser functions should only emit their own custom exceptions, that are declared in documentation.

conda_recipe_manager/parser/_node.py
```diff
@@ -208,3 +211,24 @@ def is_collection_element(self) -> bool:
...
+            bool(self.children)
+            and any(not child.list_member_flag for child in self.children if not child.is_comment())
+            and any(child.list_member_flag for child in self.children if not child.is_comment())
+        )
+        if inconsistent_list:
+            log.error("The node tree is inconsistent: a list node contains both list members and non-list members.")
+        # Return True only if all children are list members.
+        contains_list = bool(self.children) and all(
+            child.list_member_flag for child in self.children if not child.is_comment()
+        )
```
schuylermartin45: If you ditched using `any`/`all`, couldn't you have 1 for loop to handle all these concerns? Then it'd be 1 O(n) not 3 O(n) operations.

...also both of these could be `Final`.

conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -928,6 +928,25 @@ def render(self, omit_trailing_newline: bool = False) -> str:
 
         return "\n".join(lines)
 
+    def _preprocess_node_value(self, node: Node, replace_variables: bool) -> JsonType:
```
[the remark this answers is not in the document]
msentissi: I don't see `normalize_multiline_strings` called anywhere else in the code, so I'm not sure where this kind of function would be used elsewhere.

conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -939,74 +958,119 @@ def _render_object_tree(self, node: Node, replace_variables: bool, data: JsonTyp
...
 
         # Bootstrap/flatten the root-level
-        for child in self._root.children:
-            if child.is_comment():
-                continue
-            data.setdefault(cast(str, child.value), {})
+        if root_node.contains_list():
+            data: list[JsonType] = []
+        else:
+            data: dict[str, JsonType] = {}  # type: ignore
```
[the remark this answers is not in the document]
msentissi: Fixed.

conda_recipe_manager/parser/_node.py
```diff
@@ -208,3 +211,24 @@ def is_collection_element(self) -> bool:
...
+            if child.is_comment():
+                continue
+            if child.list_member_flag:
+                contains_list_member = True
+            else:
+                contains_non_list_member = True
+            if contains_list_member and contains_non_list_member:
+                log.error("The node tree is inconsistent: a list node contains both list members and non-list members.")
+                return False
+        return contains_list_member
```
schuylermartin45: So now I can see why you initially went down the O(3n) answer earlier. I very briefly started to type up a counter/bitmask way to do this, but that ends up also being messy for different reasons.

Maybe we should leave a comment at the top indicating that why we track both, so other readers have context. Either all members must be list items or none of them are. But we break as soon as we know there's a mix.

schuylermartin45 (approved):
I left 1 small request for some comment documentation to give context to other readers.

But I'm going to go ahead and approve this. Well done! Lots of good work here!

conda_recipe_manager/parser/_node.py
msentissi: Done!

PR #460, opened 2025-11-12
Recipe Rendering from Build Variants

msentissi:
### Description
The goal of this PR is to introduce variants rendering capability in CRM. This means generating a set of variant dictionaries (holding variable names and their values, this was done in #450), and using each one to render a variant of the recipe by:
- Evaluating selector expressions (done in this PR)
- Evaluating JINJA expressions (done in this PR)

### Changes
- Added `is_key()` to `Node` to check whether a node is a key. This is used in `RecipeReader` in order to detect duplicate keys in a recipe and error out during parsing.
- Added `BuildContext` class to construct and manage the dictionary that defines build environment values needed for both selector and JINJA expression evaluation.
  - This means that this class replaces and generalizes `SelectorQuery`, which was constrained to selector environment variables.
- Modified `CbcParser` to ingest all CBC variables as `str`, and use `BuildContext` instead of the obsolete `SelectorQuery`.
- Added `BuildContext` exceptions.
- Added recursive functions to `RecipeParser` to enable evaluation of selector expression and JINJA expressions.
- `RecipeReader` is modified to:
  - Check for duplicate keys (invalid YAML) during parsing as mentioned above.
  - If a JINJA variable is defined several times, consider the last definition instead of the first one for now.
  - `_render_jinja_vars` is modified to optionally accept a context for JINJA expression evaluation: this is used in `RecipeParser` to evaluate JINJA expressions for a particular recipe build variant.
- `SelectorParser` is radically simplified: `evalidate` is used to safely evaluate selector expressions instead of custom parsing/evaluation logic.
- `SelectorQuery` is completely removed as mentioned above.
- Added a `VariantsManager` class which ties everything together: given a recipe file path, and cbc file paths from which to generate build variant variables values, as well as an initial build context, it produces a base recipe (unmodified/unrendered) and all possible rendered variants.
- Tests are added/modified appropriately.
- *NOTE*: Several test files (new and existing test recipes) have been modified to remove duplicate keys.

### DIFF
--- a/conda_recipe_manager/parser/recipe_parser.py
+++ b/conda_recipe_manager/parser/recipe_parser.py
@@ -173,7 +173,7 @@ class RecipeParser(RecipeReader):
         node.comment = comment
         # Some lines of YAML correspond to multiple nodes. For consistency, we need to ensure that comments are
         # duplicate across all nodes on a line.
-        if node.is_single_key() and not node.children[0].list_member_flag:
+        if node.is_single_key():
             node.children[0].comment = comment
 
         self._rebuild_selectors()
@@ -225,7 +225,7 @@ class RecipeParser(RecipeReader):
         node.comment = new_comment
         # Some lines of YAML correspond to multiple nodes. For consistency, we need to ensure that comments are
         # duplicate across all nodes on a line.
-        if node.is_single_key() and not node.children[0].list_member_flag:
+        if node.is_single_key():
             node.children[0].comment = new_comment
 
         self._rebuild_selectors()
@@ -271,7 +271,7 @@ class RecipeParser(RecipeReader):
         node.comment = comment
         # Comments for "single key" nodes apply to both the parent and child. This is because such parent nodes render
         # on the same line as their children.
-        if node.is_single_key() and not node.children[0].list_member_flag:
+        if node.is_single_key():
             node.children[0].comment = comment
         self._is_modified = True
 


conda_recipe_manager/parser/_node.py
```diff
@@ -159,6 +159,14 @@ def is_tof_comment(self) -> bool:
         """
         return self.is_comment() and self.comment_pos == CommentPosition.TOP_OF_FILE
 
+    def is_key(self) -> bool:
```
schuylermartin45: Kind of shocked this isn't already a thing. I also don't remember why this class doesn't follow the `_<member>` naming convention, but that's an issue for another time.

environment.yaml
```diff
@@ -18,6 +18,7 @@ dependencies:
   - pip
   - click >=8.1.7,<8.2.0
   - conda
+  - evalidate >=2.0.3,<3.0.0
```
schuylermartin45: I'm still not entirely sure how I feel about adding this as a dependency. I just don't know if we, as the Conda community, have vetted this enough. I'm not saying we haven't, I just don't know. This is a small project and it is always best to _trust, but verify_ new dependencies that you add to your project.

Allegedly it comes with some kind of allow-list to filter "dangerous" code. Now I would think that wouldn't be necessary if you aren't _actually executing_ the work and are just relying on the AST's parser.

I might skim this project, but @jaimergp and @jezdez, I would be curious if you have any additional details on the vetting process before this was added as a `conda-build` dependency.

conda_recipe_manager/parser/build_context.py
```diff
@@ -0,0 +1,153 @@
...
+            python_version_int: Final[str] = self._build_env_vars["python"].replace(".", "")
+            if not python_version_int.isdigit():
+                raise BuildContextVersionException("Python", self._build_env_vars["python"])
+            context["py"] = int(python_version_int)
+            context["py3k"] = self._build_env_vars["python"].startswith("3.")
+            context["py2k"] = self._build_env_vars["python"].startswith("2.")
+            context["py27"] = context["py"] == 27
+            context["py34"] = context["py"] == 34
+            context["py35"] = context["py"] == 35
+            context["py36"] = context["py"] == 36
```
schuylermartin45: Did this nomenclature stop with 36? I could have sworn I've seen folks use this number for 3.7 and 3.8 at least? And if so, does that imply we need to manage hardcoded checks for all future versions?

conda_recipe_manager/parser/build_context.py
```diff
@@ -0,0 +1,153 @@
...
+        return context
+
+    def _construct_build_context(self) -> dict[str, Primitives]:
+        """
+        Constructs the context for the build.
+
+        :returns: The constructed build context.
+        """
+        return {
+            **self._build_env_vars,
```
schuylermartin45: Does combining these into 1 dictionary potentially risk having name collisions? Do we need to keep these categories separate?

conda_recipe_manager/parser/variants_manager.py
```diff
@@ -0,0 +1,70 @@
...
+        """
+        Returns the base recipe as a RecipeParserDeps object.
+        """
+        return self._base_recipe
+
+    def get_recipe_variants(self) -> list[RecipeReaderDeps]:
+        """
+        Returns the recipe variants as RecipeReaderDeps objects.
+        """
+        return cast(list[RecipeReaderDeps], self._recipe_variants)
```
schuylermartin45: I don't see any editing capability here. In fact, I don't think we want the "rendered variants" to be editable right? Shouldn't we just make the instances read-only to begin with(?)

schuylermartin45: Oh I guess this is an artifact of how `filter_by_selectors()` edits the tree directly. Interesting. I don't think this is how I would have done it, but I'm also not opposed to the idea.

Though....this does mean that we now allow library users a way to potentially corrupt their recipe files in a way that can't undo. Maybe as a middle-ground we make another subclass parser that gives access to these more dangerous operations(?)

schuylermartin45: To put another way: up until the introduction of these two new parser functions (`filter_by_selectors()` and `evaluate_jinja_expressions()`), users could _in theory_ always end up back with a recipe file that can write back to disk. These two new functions mean they could potentially write a _rendered variant_ recipe to disk. I feel like that we should highlight that behavior.

conda_recipe_manager/parser/recipe_parser_deps.py
```diff
@@ -195,16 +195,18 @@ def add_dependency(
...
         if patch_success and dep.selector is not None:
             sel_path = patch_path
             # `add_selector()`, by nature of how selectors work, does not support "append" mode. If an append operation
             # occurred, we must calculate the position of the last array element. We only add selectors on a successful
-            # patch, so we know we can make assume a dependency list exists.
+            # patch, so we know we can assume a dependency list exists.
+            # If we are adding a new section, we need to add the selector to the first dependency.
             if sel_path.endswith("/-"):
                 sel_path = sel_path[0:-1] + str(len(cast(list[str], self.get_value(base_path))) - 1)
+            elif is_new_section:
```
schuylermartin45: This feels like a good edge-case catch to me!

tests/parser/test_variants_manager.py
```diff
@@ -0,0 +1,72 @@
...
+        cbc_paths.append(recipe_cbc_path)
+
+    manager = VariantsManager(
+        recipe_path=recipe_path,
+        cbc_paths=cbc_paths,
+        build_context=BuildContext(platform=platform),
+    )
+
+    for i, variant in enumerate(manager.get_recipe_variants()):
+        assert variant.render() == load_file(f"recipe_variants/{feedstock}/{platform}_{i}.yaml")
```
schuylermartin45: I don't remember where I made that comment/can't currently find it, but I now see the reason for numbering the files. That makes perfect sense, so please just resolve that comment.

conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -654,10 +663,43 @@ def _private_init(
             cur_indent = new_indent
             # Look at the stack to determine the parent Node and then append the current node to the new parent.
             parent = node_stack[-1]
+            # Check for duplicate keys and bail if found.
+            if new_node.is_key() and not new_node.list_member_flag:
```
schuylermartin45: I would just combine these two into 1 if

conda_recipe_manager/parser/cbc_parser.py
```diff
@@ -144,7 +144,9 @@ def __init__(self, content: str):
                 path = f"/{variable}/{i}"
                 if is_single_value:
                     path = f"/{variable}"
-                entry = self._construct_cbc_variable(path, value, comments_tbl)
+                # This is necessary to ensure closeness with conda-build's format.
```
schuylermartin45:
```suggestion
                # This is necessary to ensure alignment with conda-build's format.
```
Maybe this sounds better?

conda_recipe_manager/parser/selector_parser.py
```diff
@@ -5,104 +5,79 @@
...
+        """
+        Prepares the EvalModel for the selector parser.
+        The model is setup similarly to conda-build's in order to support the same set of expressions.
+
+        :returns: The prepared EvalModel.
+        """
+        model: EvalModel = base_eval_model.clone()  # type: ignore[no-any-unimported, misc]
+        model.nodes += ["Call", "Attribute", "Tuple", "List", "Dict", "Is", "IsNot"]  # type: ignore[misc]
+        model.allowed_functions += ["int", "float", "len", "str", "list", "dict", "tuple"]  # type: ignore[misc]
+        model.attributes += [  # type: ignore[misc]
```
schuylermartin45: This is definitely an interesting idea. But at the same time, in an ideal world, we shouldn't need to allow-list functionality. If we are using the AST library correctly, we should be able to interpret the parse tree correctly without _executing_ Python code in selectors, right(?)

This to me implies a full evaluation is being done at _some point_. But admittedly I still haven't had the time to read this library yet.

Of course in an ideal ideal world we'd have a grammar that fully defines what selectors can do in V0 recipe files.

....and in a really really ideal world we wouldn't be dealing with a recipe format that shipped to allow Python code execution in comments but here we are....I guess we're the [desperate measures](https://www.youtube.com/watch?v=u0VjwHY81lo&t=197s).

conda_recipe_manager/parser/selector_parser.py
```diff
@@ -328,46 +113,33 @@ def __eq__(self, other: object) -> bool:
...
-                            return _eval_node(node.l_node) or _eval_node(node.r_node)
-                case str():
-                    return node.value in query.build_env_vars
 
-        return _eval_node(self._root)
+        selector_context: Final[dict[str, Primitives]] = build_context.get_selector_context()
+
+        try:
+            expr_code = Expr(self._content, model=SelectorParser._get_evalidate_model()).code  # type: ignore[misc]
+            while True:
```
schuylermartin45: This forever loop that tries/retries to evaluate code scares me a bit. There's got to be another way, right?

conda_recipe_manager/parser/selector_parser.py
```diff
@@ -5,104 +5,79 @@
...
-SelectorNodeNestedList = list[Union[_SelectorNode, "SelectorNodeNestedList"]]
-
-
-class SelectorParser(IsModifiable):
-    """
-    Parses a selector statement
-    """
+        # Sanitize content string
+        # TODO Future: validate with Selector regex for consistency, not string indexing.
+        if self._schema_version == SchemaVersion.V0 and content and content[0] == "[" and content[-1] == "]":
```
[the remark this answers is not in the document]
msentissi: I'd prefer to merge this PR sooner rather than later.

conda_recipe_manager/parser/selector_parser.py
```diff
@@ -328,46 +113,33 @@ def __eq__(self, other: object) -> bool:
...
-                    return node.value in query.build_env_vars
 
-        return _eval_node(self._root)
+        selector_context: Final[dict[str, Primitives]] = build_context.get_selector_context()
+
+        try:
+            expr_code = Expr(self._content, model=SelectorParser._get_evalidate_model()).code  # type: ignore[misc]
+            while True:
+                try:
+                    # expr_code is already guaranteed to be safe to evaluate
```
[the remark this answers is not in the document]
msentissi: It's *meant* to be guaranteed to be safe to evaluate

conda_recipe_manager/parser/variants_manager.py
```diff
@@ -0,0 +1,70 @@
...
+class VariantsManager:
+    """
+    Class that manages the variants of a recipe, given a list of CBC files.
+    """
+
+    def __init__(self, recipe_path: Path, cbc_paths: list[Path], build_context: BuildContext):
+        """
+        Initializes the VariantsManager.
+
+        :param recipe_path: Path to the recipe file.
```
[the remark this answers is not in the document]
msentissi: Fixed!

tests/parser/test_recipe_reader.py
```diff
@@ -1646,3 +1646,12 @@ def test_unsupported_jinja2_statements_parsing(
...
     parser = load_recipe(file, RecipeReader, force_remove_jinja)
     assert parser.render() == load_file(rendered_file)
+
+
+def test_duplicate_keys() -> None:
+    """
+    Tests that the recipe reader correctly handles duplicate keys.
+    """
+    with pytest.raises(DuplicateKeyException) as e:
+        load_recipe("duplicate_keys/google-cloud-cpp.yaml", RecipeReader)
```
schuylermartin45: Nitpick: Should we make this a parameterized test to make it easier to add more test files in the future (in case of a regression)?

conda_recipe_manager/parser/recipe_variant.py
```diff
@@ -0,0 +1,111 @@
+"""
+:Description: Provides a class that represents a variant of a recipe.
+
+This is a subclass of RecipeReaderDeps whose constructor takes a BuildContext object and uses it to construct
+a rendered variant of the recipe instead of the recipe in its original form.
+It is intended to give full read access to the recipe variant.
+
+This class is intended to be used in the context of the VariantsManager class.
```
schuylermartin45: I like this description! Good job!

conda_recipe_manager/parser/recipe_variant.py
```diff
@@ -0,0 +1,111 @@
...
+                    continue
+                _filter_selectors_and_paths(child)
+            node.children = new_children
+
+        _filter_selectors_and_paths(self._root)
+
+        self._rebuild_selectors()
+        self._is_modified = True
+
+    def _evaluate_jinja_expressions(self, build_context: BuildContext) -> None:
```
schuylermartin45: Wait, don't we want to make these public now that they are in a subclass?

conda_recipe_manager/parser/recipe_parser.py
```diff
@@ -173,12 +174,35 @@ def _extract_selector(s: str) -> str:
         node.comment = comment
         # Some lines of YAML correspond to multiple nodes. For consistency, we need to ensure that comments are
         # duplicate across all nodes on a line.
-        if node.is_single_key():
+        if node.is_single_key() and not node.children[0].list_member_flag:
```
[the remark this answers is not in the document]
msentissi: Undone!

PR #465, opened 2025-12-12
Changelog correction

msentissi:
### Description

A slight changelog correction.


msentissi:
> I'm not sure why we suddenly are getting a redundant cast warning. Did we take in a new mypy version?

Seems to be the case, I updated locally and now I'm matching the CI pre-commit results.

PR #471, opened 2026-01-19
only render JINJA variables when necessary

msentissi:
### Description

All node values were being processed through `self._render_jinja_vars(node.value, context)`. We now check whether the node value contains a JINJA expression first, and perform the rendering conditioned on that.


msentissi:
> Do you want to add any tests related to this change? Like a regression check maybe?

Yes, done!

PR #499, opened 2026-03-18
allow duplicate keys flag

leahsch:
- Adds `RecipeReaderFlags.ALLOW_DUPLICATE_KEYS` to allow duplicate YAML keys in a recipe file if desired.
  - If enabled, allows recipe parsing to continue even if duplicate keys are encountered. This prints a warning to logs as well.
- Calling code may enable a `warnings.warn()` if duplicate keys are allowed, which will propagate the warning up to the calling code to be caught. This allows calling code to do whatever is necessary with the warning, without interrupting the flow of code execution as an exception would.
    - By default the warning is disabled, so should not interfere with any normal use of CRM.
- Small refactor of some test/test helper code to make testing with `RecipeReaderFlags` more straightforward.

### DIFF
--- a/conda_recipe_manager/parser/recipe_reader.py
+++ b/conda_recipe_manager/parser/recipe_reader.py
@@ -816,7 +816,9 @@ class RecipeReader(IsModifiable):
                 # This warning is disabled in conda-recipe-manager by default and up to the client to enable if needed.
                 # If enabled, warnings.warn will raise the warning to any calling code to be caught without
                 # interrupting code execution like an exception would. This will print to stderr as well.
-                # See /conda-recipe-manager/__init__.py for an example of enabling it.
+                # See:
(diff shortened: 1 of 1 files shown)


isuruf:
@leahsch with this PR merged, I can't convert v0 recipes using `crm convert` that have duplicate keys (eg: `script` keys with win and unix selectors). Is there a way to allow those?

schuylermartin45:
We were pretty careful to try to preserve the original functionality as to not to disturb users. Apologies if we missed something. We also haven't noticed a compatibility drop in `crm convert` integration tests around these changes.

Hopefully your issue could be solved by either exposing the `ALLOW_DUPLICATE_KEY` flag to the `crm convert` CLI or by (ideally) changing the default flags used in the `crm convert` command.

PR #504, opened 2026-04-03
Fix parser crash on compact nested list syntax in CBC files

psteyer:
### Description

Fix `_render_object_tree` crash on compact nested list syntax in CBC files.

When a `conda_build_config.yaml` uses the compact nested list form for `zip_keys`:

```yaml
zip_keys:
  - - target_machine
    - cross_target_platform
    - centos_machine
```  

The parser crashes with `AttributeError: 'list' object has no attribute 'setdefault'`. 
This is because `_parse_line_node` stores the inner list `(["target_machine"])` as the node's value, which causes `_render_object_tree` to call `dict.setdefault()` on a list object.


tests/test_aux_files/cbc_files/zip_keys_compact_nested_list_rendered.yaml
```diff
@@ -0,0 +1,19 @@
...
+  - linux-64
+  - linux-aarch64
+
+ctng_vendor:
+  - conda
+
+zip_keys:
+    - target_machine
+    - cross_target_platform
+    - centos_machine
```
schuylermartin45: This is just a single-level list. Shouldn't it be:
```
zip_keys:
  -
    - target_machine
    - cross_target_platform
    - centos_machine
```
?

conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -463,6 +463,22 @@ def _parse_line_node(s: str, only_seen_comments: bool, yaml_loader: type[SafeLoa
...
+        becomes:
+          "  -"
+          "    - target_machine"
+
+        :param yaml_str: Raw YAML string.
+        :returns: YAML string with compact nested lists expanded.
+        """
+        if not yaml_str:
+            return yaml_str
+        return re.sub(r"^(\s*)- - ", r"\1-\n\1  - ", yaml_str, flags=re.MULTILINE)
```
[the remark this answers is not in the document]
psteyer: I think the regex handles the extra space between the `-`s now.

I am wondering if it's worth expanding it too also be able to handle a use case where there's an extra spaces in the inner items?

conda_recipe_manager/parser/v0_recipe_formatter.py
```diff
@@ -54,6 +55,20 @@ def is_v0_recipe(self) -> bool:
         """
         return self._is_v0_recipe
 
+    def expand_compact_nested_lists(self) -> None:
+        """
+        Expands compact nested list syntax into the equivalent expanded form. For example:
+          "  - - target_machine"
+        becomes:
+          "  -"
+          "    - target_machine"
```
schuylermartin45, line 64: The quotes are a little clunky but it gets the point across. You could use ``` in the comments I guess(?)

Not a big deal, just spitballing here.

conda_recipe_manager/parser/v0_recipe_formatter.py
```diff
@@ -54,6 +55,20 @@ def is_v0_recipe(self) -> bool:
...
+
+        Modifies ``self._lines`` in place.
+        """
+        text = "\n".join(self._lines)
+        text = re.sub(r"^([ \t]*)-[ \t]+-[ \t]+", r"\1-\n\1  - ", text, flags=re.MULTILINE)
+        self._lines = text.splitlines()
```
schuylermartin45, line 70: Merging the text to then immediately split it back up again seems like a very memory-intensive operation. I get that this was done for the sake of running `re.sub()` but there should be a better way here.

Also by convention, all of the regular expressions in CRM are maintained in the `Regex` class (which is really just being treated as a C++ish `namespace` to group all the regular expressions together). All of the regular expressions are defined with `re.compile()` for performance reasons. Allegedly the C-Python interpreter will cache regular expressions for us anyways, but I don't want to be in the habit of making assumptions based on the interpreter implementation.

psteyer: I moved the regex over to `_types`, and tried to follow a similar pattern.

Then instead of doing the splitting and then merge pattern, I tried to just go line by line.

conda_recipe_manager/parser/_types.py
```diff
@@ -197,6 +197,9 @@ class Regex:
     PRE_PROCESS_MIN_PIN_REPLACEMENT: Final[re.Pattern[str]] = re.compile(r"min_pin=")
     PRE_PROCESS_MAX_PIN_REPLACEMENT: Final[re.Pattern[str]] = re.compile(r"max_pin=")
 
+    # Attempts to normalize compact nested list syntax into the equivalent expanded form.
+    PRE_PROCESS_COMPACT_NESTED_LIST: Final[re.Pattern[str]] = re.compile(r"^([ \t]*)-[ \t]+-[ \t]+", flags=re.MULTILINE)
```
schuylermartin45: I don't think `re.MULTILINE` is needed if you use this line-by-line in a loop below.

schuylermartin45:
```suggestion
    PRE_PROCESS_COMPACT_NESTED_LIST: Final[re.Pattern[str]] = re.compile(r"^([ \t]*)-[ \t]+-[ \t]+")
```

schuylermartin45 (approved):

Issue #506, opened 2026-04-07
`CbcParser` should follow `Reader`/`Parser` nomenclature


schuylermartin45:
This is a minor consistency issue that hopefully can be done without breaking downstream users. `CbcParser` should actually be `CbcReader` and then we can reserve `CbcParser` for editing capabilities.

PR #513, opened 2026-04-09
Fixes Issue #506, refactors `CbcParser`

schuylermartin45:
- `CbcParser` --> `CbcReader` to follow in the conventions established by the `Recipe*` classes
- `CbcParser` still exists for compatibility purposes AND for future potential feature development
- Moves `CbcOutputType` and `GeneratedVariantsType` under `parser.types` instead of the `CbcParser` class.
- Improves documentation accordingly


tests/parser/test_cbc_reader.py
```diff
@@ -236,9 +234,9 @@ def test_get_cbc_variable_values_raises(
     :param build_context: Target build context
     :param exception: Exception expected to be raised
     """
-    parser = load_cbc(file)
+    reader: Final = load_cbc(file)
```
leahsch, line 237: is there a reason this test got assigned to a variable but the others are just left as `load_cbc(file)`?
I guess because you don't want `load_cbc()` called in the `.raises()` block?

schuylermartin45: Yup, exactly that!

leahsch (approved):

PR #517, opened 2026-04-22
Allow duplicate keys in convert command

isuruf:
### Description

Since v0 recipes can have duplicate keys with preprocessing selectors, it's good to enable this option by default.

### DIFF
--- a/conda_recipe_manager/commands/convert.py
+++ b/conda_recipe_manager/commands/convert.py
@@ -138,7 +138,7 @@ def convert_file(
         )
 
     # Parse the recipe
-    base_flags = RecipeReaderFlags.NONE
+    base_flags = RecipeReaderFlags.ALLOW_DUPLICATE_KEYS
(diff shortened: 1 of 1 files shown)


schuylermartin45 (approved):
I worry this may introduce a bunch of "silent" conversion issues.

Duplicate keys break the `patch()` system, which closely follows the JSON Patch RFC (but ya know, for YAML not JSON). 

If there are two identical paths, I _believe_ only 1 gets patched. If that happens on a mission-critical key in the upgrade process, it could cause some very nasty bugs.

That being said, this 1 change bumps our conversion success rate from 97% to 99% in the [Integration Tests](https://github.com/conda/conda-recipe-manager/actions/runs/24803979395?pr=517).

If it weren't for the integration test difference, I would probably push against this a bit more. I guess we'll just have to see how many recipes fall into this camp, or if most of these conversions just duplicate something untouched, like the `/build/number` key.

PR #519, opened 2026-04-23
Upgrades to the new (and faster) setup-miniconda action

schuylermartin45:
Gotta go fast.


jezdez:
Doesn't seem to do much :-/

schuylermartin45:
> Doesn't seem to do much :-/

I saw a 10s improvement in a few spots, but yeah, not a huge jump.

That being said, I'd still like to have this as a pinned dependency going forward.

leahsch (approved):

PR #530, opened 2026-05-15
Recover when a list-leaf is rendered into a dict context

ericdill:
## Summary

CBCs that pin a key only on a particular platform are sometimes written as an empty key with a selector comment, immediately followed by a list item also under a selector. For example:

```yaml
c_compiler:                    # [win]
- vs2019                       # [win]
cxx_compiler:                  # [win]
- vs2019                       # [win]
```

The parser promotes the list item (`vs2019`) to a sibling of the empty-key (`c_compiler`) instead of nesting it underneath. When `_render_object_tree` then encounters the list leaf with the root dict as its `data` argument, it crashes:

```
File ".../conda_recipe_manager/parser/recipe_reader.py", line 1254, in _render_object_tree
    data.append(self._preprocess_node_value(node, replace_variables))
AttributeError: 'dict' object has no attribute 'append'
```

The comment on the offending line (`# At this point, we know the data is a list`) makes the assumption explicit, which is exactly the assumption that breaks here.

## Fix

Recover at the render layer: when a strong-leaf node arrives with a dict parent, attach it back under the most recent empty-key sibling — the only intent that makes sense for this YAML shape.

The deeper fix is at the parser level (don't let the empty-key's children get reparented when both have selector comments), but that's a much larger change with broader risk. This unblocks downstream consumers without changing the success path. Happy to take a different direction if the maintainers prefer the deeper fix.

## Test plan

- [x] Added a regression test in `tests/parser/test_cbc_parser.py` with a 4-line repro
- [x] Confirmed the new test passes
- [x] All 722 tests in `tests/parser/` pass with the change

## Discovery context

Found while debugging a PBP linter failure on [AnacondaRecipes/tesseract-feedstock#6](https://github.com/AnacondaRecipes/tesseract-feedstock/pull/6). The crash blocks anaconda-linter against any feedstock whose `recipe/conda_build_config.yaml` happens to take this YAML shape — the original tesseract recipe had it from a 2022-era snapshot of `vs2019` Windows pins.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/conda_recipe_manager/parser/recipe_reader.py
+++ b/conda_recipe_manager/parser/recipe_reader.py
@@ -1218,8 +1218,31 @@ class RecipeReader(IsModifiable):
             return
 
         if node.is_strong_leaf():
-            # At this point, we know the data is a list
-            data.append(self._preprocess_node_value(node, replace_variables))
+            value = self._preprocess_node_value(node, replace_variables)
+            # The common case: a strong leaf appears as a list member, so `data` is the
+            # parent list to which we append. If the parser instead handed us a dict, the
+            # leaf was promoted to a sibling of an empty-key node — typically because of
+            # selector comments / blank lines disturbing list nesting (see
+            # https://github.com/conda/conda-recipe-manager/issues/<TBD>). Recover by
+            # attaching the leaf as a single-item list under the most recent empty-key
+            # sibling, which preserves the data the user clearly intended.
+            if isinstance(data, list):
+                data.append(value)
+            elif isinstance(data, dict) and data:
+                last_key = next(reversed(data))
+                if data[last_key] is None:
+                    data[last_key] = [value]
+                elif isinstance(data[last_key], list):
+                    data[last_key].append(value)
+                else:
+                    raise TypeError(
+                        f"List leaf {value!r} cannot be attached: previous key "
+                        f"{last_key!r} already has a non-list value."
+                    )
+            else:
+                raise TypeError(
+                    f"List leaf {value!r} appeared with no list context to append to."
+                )
             return
 
         # Process collection nodes (lists or dicts that are list members)
--- a/tests/parser/test_cbc_parser.py
+++ b/tests/parser/test_cbc_parser.py
@@ -565,3 +565,25 @@ def test_generate_variants(
         assert _find_matching_variant(exp_var, generated_variants)
     for gen_var in generated_variants:
         assert _find_matching_variant(gen_var, expected_variants)
+
+
+def test_empty_key_with_selector_followed_by_list() -> None:
+    """
+    Regression test: CBCs that pin a key only on a particular platform are sometimes
+    written as an empty key with a selector comment, immediately followed by a list
+    item also under a selector. This shape used to crash `_render_object_tree` with
+    "'dict' object has no attribute 'append'" because the parser promoted the list
+    item to a sibling of the empty key. The render layer now recovers by attaching
+    the leaf back under the most recent empty-key sibling.
+    """
+    cbc = (
+        "c_compiler:                    # [win]\n"
+        "- vs2019                       # [win]\n"
+        "cxx_compiler:                  # [win]\n"
+        "- vs2019                       # [win]\n"
+    )
+    parser = CbcParser(cbc)
+    assert parser.get_value("/") == {
+        "c_compiler": ["vs2019"],
+        "cxx_compiler": ["vs2019"],
+    }


tests/parser/test_cbc_parser.py
```diff
@@ -569,3 +565,25 @@ def test_generate_variants(
...
+        "c_compiler:                    # [win]\n"
+        "- vs2019                       # [win]\n"
+        "cxx_compiler:                  # [win]\n"
+        "- vs2019                       # [win]\n"
+    )
+    parser = CbcParser(cbc)
+    assert parser.get_value("/") == {
+        "c_compiler": ["vs2019"],
+        "cxx_compiler": ["vs2019"],
+    }
```
schuylermartin45, line 589: Given the sheer number of scenarios that have to cover with these parsers, we're big proponents of using `pytest.mark.parametrize()` in CRM over 1-off custom tests. In our experience it also makes it much easier to maintain so many tests over time.

This should be able to be re-worked as a parametrized case of `test_generate_variants()` above, right? Just copy the contents of the multi-line string into a file and fill out the settings. That also allows other future tests to use that example CBC file without having to hard-code the string.

conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -1250,8 +1218,31 @@ def _render_object_tree(self, node: Node, replace_variables: bool, data: JsonTyp
...
 
         if node.is_strong_leaf():
-            # At this point, we know the data is a list
-            data.append(self._preprocess_node_value(node, replace_variables))
+            value = self._preprocess_node_value(node, replace_variables)
+            # The common case: a strong leaf appears as a list member, so `data` is the
+            # parent list to which we append. If the parser instead handed us a dict, the
+            # leaf was promoted to a sibling of an empty-key node — typically because of
+            # selector comments / blank lines disturbing list nesting (see
+            # https://github.com/conda/conda-recipe-manager/issues/<TBD>). Recover by
```
schuylermartin45, line 1226: I assume the `<TBD>` is mean to be a reference to this PR?

conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -1250,8 +1218,31 @@ def _render_object_tree(self, node: Node, replace_variables: bool, data: JsonTyp
...
+                last_key = next(reversed(data))
+                if data[last_key] is None:
+                    data[last_key] = [value]
+                elif isinstance(data[last_key], list):
+                    data[last_key].append(value)
+                else:
+                    raise TypeError(
+                        f"List leaf {value!r} cannot be attached: previous key "
+                        f"{last_key!r} already has a non-list value."
+                    )
```
schuylermartin45, line 1241:
```suggestion
                raise TypeError(
                    f"List leaf {value} cannot be attached: previous key "
                    f"{last_key} already has a non-list value."
                )
```
Dangling else is not necessary. CRM prefers the return-early paradigm when possible. CRM also tends to prefer rendering with `str()` over `repl()` so getting rid of `!r` would be more consistent with the current library.

conda_recipe_manager/parser/recipe_reader.py
schuylermartin45, line 1195: Missing docstring: with this change we also throw a `TypeError`

schuylermartin45: Presumably we should also note that in all other functions that currently invoke and do not catch `TypeError`.

schuylermartin45: ...which then raises the question, what are those functions and how big of a deal is that we can throw a `TypeError()` now?

conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -1250,8 +1218,31 @@ def _render_object_tree(self, node: Node, replace_variables: bool, data: JsonTyp
...
+            # attaching the leaf as a single-item list under the most recent empty-key
+            # sibling, which preserves the data the user clearly intended.
+            if isinstance(data, list):
+                data.append(value)
+            elif isinstance(data, dict) and data:
+                last_key = next(reversed(data))
+                if data[last_key] is None:
+                    data[last_key] = [value]
+                elif isinstance(data[last_key], list):
+                    data[last_key].append(value)
```
schuylermartin45, line 1236: The more I think of this, the more something doesn't seem right to me about these changes. In general, nodes aren't supposed to contain dictionary or list values. There are exceptions and off the top of my head I want to say multi-line strings are represented as lists under one parse tree node.

I guess what I'm trying to say is that this smells like there's a corruption in the parse tree on load and we're fixing it with a render-time patch. If that's the case, it is probably more apparent in the "debug" view of the parse tree. 

The way I see the debug tree is either run `crm convert --debug <file>` and look at the V0 version of the parse tree when the parser first reads the recipe. Or I pull out the Python interpeter and do something like:
```py
from pathlib import Path
from conda_recipe_manager.parser.recipe_reader import RecipeReader
print(str(RecipeReader(Path("/path/to/recipe").read_text())))
```

Now granted, it has been a while since I've addressed a CRM parsing issue like this and maybe I'm forgetting something.

conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -1250,8 +1218,31 @@ def _render_object_tree(self, node: Node, replace_variables: bool, data: JsonTyp
...
+                else:
+                    raise TypeError(
+                        f"List leaf {value!r} cannot be attached: previous key "
+                        f"{last_key!r} already has a non-list value."
+                    )
+            else:
+                raise TypeError(
+                    f"List leaf {value!r} appeared with no list context to append to."
+                )
             return
```
schuylermartin45, line 1246: Am I reading this right? Or is this a quirk of the GitHub view I have opened?

This return-early-paradigm version would be logically equivalent, right?
```suggestion
            raise TypeError(
                f"List leaf {value} appeared with no list context to append to."
            )
```
And that would mean if we didn't hit any branching path covered under "strong leaf" then we throw. That feels like a pretty dramatic shift in behavior.

schuylermartin45 (commented):
I think the LLM picked up on this nuance by some of its comments in the PR body, but I suspect this is patching a parse time issue at render-time. I could be wrong, but I think the real issue is that the parse tree is reading the file incorrectly in this particular set of circumstances.

I have a write-up on explaining my thinking around the big `if isinstance()` branching.

Thanks for the PR and highlighting this issue. We definitely have to fix something lol

ericdill:
thanks for the deep review @schuylermartin45 . am i understanding your overall position correctly as "lets not patch this on the render-time side since this looks to be a parse-time issue"?

i'll look into what it would look like to flag this on the parse-time side of things.

ericdill:
so with this as the conda-build-config.yaml:
```yaml
c_compiler:                    # [win]
- vs2019                       # [win]
cxx_compiler:                  # [win]
- vs2019                       # [win]
```

crm debug produces:
```
  /
    |- c_compiler
    |- vs2019
    |- cxx_compiler
    |- vs2019
```

its putting everything at the same level. pyyaml handles this correctly. yaml.safe_load returns vs2019 as children of c_compiler and cxx_compiler. CRM's parser only pushes the stack when new_indent > cur_indent, never with new_indent == cur_indent, even if the previous node was a key expecting a list child. 

So, the proposed parse_time fix would be: when seeing a list-member at equal indent immediately after a key with no children, push the key onto the stack so the list-member is nested underneath it.

roughly this change
```python
  # Existing:
  if new_indent > cur_indent:
      node_stack.append(last_node)
  # New: equal-indent list-member-after-key case
  elif (
      new_indent == cur_indent
      and new_node.list_member_flag
      and last_node.is_key()
      and not last_node.children  # last_node has no value yet
  ):
      node_stack.append(last_node)
```

i can have my agent do this and check / fix the test suite where anything breaks. but this feels like a maybe substantial change to the parser? im not that confident in my understanding of where this might break things if users are relying on this behavior? wdyt? or maybe im over-thinking this and is hould just let claude spin on this and see what shape of a solution it comes back with.

ericdill:
Your read was right — the parse tree is corrupted at load time, not at render time. `crm convert --debug` confirms it: four flat siblings under root instead of two keys-with-children.

I've put up a parse-time fix as a draft alternative in #531. Let me know if you'd like to review that one instead and close this; happy to do whichever you prefer.

schuylermartin45:
> thanks for the deep review @schuylermartin45 . am i understanding your overall position correctly as "lets not patch this on the render-time side since this looks to be a parse-time issue"?

Yes that's a good summary. Sorry if I wasn't clear on that.

I'm not going to quote the second reply because it is huge but my reaction is: oh crap. I did not notice the lack of indentation in the previous examples. I appreciate you taking the time to find that. Had I noticed the lack of indentation I probably would have picked up on that being the issue, so apologies for not noticing sooner.

For some historical context, the parser is heavily depends on indentation levels. Given the original target was "V0 recipe files" and +90-95% of Conda recipes out in the community are well-indented this was an early cost-cutting measure. Parsing CBC files using the recipe parser was a quick hack that came much later.

Now that being said, over time we have found cases of inconsistent indentation that cause strange parser screw-ups. Instead of re-writing the V0 parser we created the `V0RecipeFormatter` to "correctly" indent the string before the parser triggers. Is this ideal: no, but it has the added benefit of linting the shape of recipe files to conform with common practices.

If this CBC file works as expected:
```yaml
c_compiler:                    # [win]
  - vs2019                       # [win]
cxx_compiler:                  # [win]
  - vs2019                       # [win]
```
then I'm perfectly fine if we follow past precedent(s) and make another `V0RecipeFormatter` indentation hack. Again it is not perfect or efficient, but it works well for the needs of our tools and the Conda Community.

ericdill:
superseded by #531 531

PR #531, opened 2026-05-15
Parse list items that share indentation with their parent key

ericdill:
## Summary

Alternative to #530 that addresses @schuylermartin45's review: this is a parse-time fix instead of a render-time one.

## Diagnosis

`crm convert --debug` on the repro CBC produced this tree:

```
/
  |- c_compiler
  |- vs2019
  |- cxx_compiler
  |- vs2019
```

The four entries are siblings under root. They should be:

```
/
  |- c_compiler
    |- vs2019
  |- cxx_compiler
    |- vs2019
```

PyYAML reads the same input as `{c_compiler: [vs2019], cxx_compiler: [vs2019]}`, confirming the YAML is valid and our parser is wrong here.

The cause is in `_init_parse_tree`: the indent-comparison only pushes the stack on `new_indent > cur_indent`. With list items at the same column as their parent key (`new_indent == cur_indent`, both 0), the push never happens.

## Fix

Two new equal-indent branches in `_init_parse_tree`:

1. When the new node is a list member, the previous node was a value-less key, and they share indent → push the key onto the stack so the list member nests underneath.
2. When we then leave that same-indent list context (next non-list sibling at the same indent) → pop the key back off.

A small `same_indent_list_active` flag tracks whether the most recent push was via case 1, so the case-2 pop only fires for pushes we made (not for normal indented lists ending naturally).

Total parser change: 28 lines (mostly comments and the two new elif branches).

## Tests

Added a minimal regression CBC fixture (`zero_indent_list_cbc.yaml`) and parametrized cases under existing tests:

- `test_list_cbc_variables` — confirms the parser sees `c_compiler` and `cxx_compiler` as the two top-level keys (not 4 flat siblings).
- `test_get_cbc_variable_values` — confirms each key's list value resolves to `["vs2019"]` for `Platform.WIN_64`.

All 724 parser tests pass (existing 721 + 3 new).

## Discovery context

Found while debugging a PBP linter failure on [AnacondaRecipes/tesseract-feedstock#6](https://github.com/AnacondaRecipes/tesseract-feedstock/pull/6) — older feedstocks have 2022-era Windows-only `vs2019` pins written in this shape and they've been silently breaking anaconda-linter's CRM-backed pipeline on every recent PR.

## Relationship to #530

This PR replaces #530. #530 caught the symptom at the render layer; this catches the cause at the parse layer. Marking this draft for review — happy to close #530 outright once you've taken a pass at this one.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/commit_checks.yaml
+++ b/.github/workflows/commit_checks.yaml
@@ -47,8 +47,11 @@ jobs:
           run: |
             source $CONDA/bin/activate
             conda create --name build-env -y python=3.11 conda-build
-            conda activate build-env
-            conda build -c defaults recipe/meta.yaml
+            # Use `conda run -n build-env` so the `conda` that picks up
+            # conda-build's plugin entry points is the one inside the env.
+            # `conda activate` alone leaves the base conda on PATH, which no
+            # longer ships `conda build` as a builtin subcommand.
+            conda run -n build-env conda build -c defaults recipe/meta.yaml
   # Eat our own dog food and build this project with rattler-build by converting our existing recipe.
   # Remember, `rattler-build` will pull from the `conda-forge` channel, which may cause some version inconsistencies.
   # NOTE: As of 2026-02-25, we tell `rattler-build` to ignore 3.14+ explicitly (via `sed`) to avoid issues with the
--- a/.github/workflows/commit_checks.yaml
+++ b/.github/workflows/commit_checks.yaml
@@ -47,11 +47,11 @@ jobs:
           run: |
             source $CONDA/bin/activate
             conda create --name build-env -y python=3.11 conda-build
-            # Use `conda run -n build-env` so the `conda` that picks up
-            # conda-build's plugin entry points is the one inside the env.
-            # `conda activate` alone leaves the base conda on PATH, which no
-            # longer ships `conda build` as a builtin subcommand.
-            conda run -n build-env conda build -c defaults recipe/meta.yaml
+            # Call the `conda-build` binary directly. Recent conda versions
+            # don't register `conda build` as a subcommand, so `conda build`
+            # fails with `invalid choice: 'build'` whether invoked via
+            # `conda activate` or `conda run`.
+            conda run -n build-env conda-build -c defaults recipe/meta.yaml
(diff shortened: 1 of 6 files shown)


schuylermartin45:
I posted my reply on the other PR before I saw this but if this works, it is arguably a better fix than the `V0RecipeFormatter` change I initially suggested. 

At first glance it looks right and reads pretty good for how nasty the existing branching logic is.

My primary suggestion is that maybe we add some parser "gotcha" tests to the `RecipeReader` tests to see how well this works in lists of lists or lists of objects.

I also have to leave a bit early today so I may not get to reviewing the non-draft PR until Monday.

ericdill:
Thanks!

> My primary suggestion is that maybe we add some parser "gotcha" tests to the RecipeReader tests to see how well this works in lists of lists or lists of objects.

Done in 756636e. Both shapes work as-is — the change only adjusts when value-less keys get pushed onto the stack, and the existing collection-element logic continues to do the right thing once the nesting is correct. Added two fixtures and a direct `get_value("/")` assertion covering simple lists, list-of-lists, and list-of-objects. Verified the parsed structure matches PyYAML's `safe_load` output for each.

Also fixed the formatting issue that was tripping pre-commit (collapsed two elif blocks onto single lines, line length fits at 120). Pylint clean. Take your time on the review — Monday is fine.

ericdill:
@schuylermartin45 just a heads up i was getting a weird issue with `conda build` producing a `build subcommand not found` issue whereas if i swap the gha to `conda-build` then everything works fine. otherwise CI is green and happy

schuylermartin45:
> @schuylermartin45 just a heads up i was getting a weird issue with `conda build` producing a `build subcommand not found` issue whereas if i swap the gha to `conda-build` then everything works fine. otherwise CI is green and happy

I think this is tied to the new version of `conda` that was released late last week. It is happening on the conda-bot template PR too: https://github.com/conda/conda-recipe-manager/pull/533 so thanks for taking care of that!

schuylermartin45 (approved):

schuylermartin45:
Thanks for the contributions and for adding the extra tests!

PR #536, opened 2026-06-12
Fixes a bad cast and a new `black` format warning

schuylermartin45:
- Fixes a bad `cast` that incorrectly suppressed a legitimate `mypy` error. This caused the wrong exception to be thrown when a package failed to be acquired.
- Also fixes a newer black formatting warning


schuylermartin45:
Normally I'd aim to put in a regression test but I can't think of anything super meaningful here. This was just an incredibly bad `cast` that should not have been made.

leahsch (approved):

PR #539, opened 2026-06-23
Fix detect binary files prefix inversion

chrisburr:
### Description

In https://github.com/conda-forge/texlive-core-feedstock/pull/110 `conda-recipe-manager` incorrectly converted the recipe due to the boolean flag for `detect_binary_files_with_prefix` not being inverted when it was renamed to `ignore_binary_files`.

This PR adds a test and fixes it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ Note: version releases in the 0.x.y range may introduce breaking changes.
 ### Deprecated
 ### Removed
 ### Fixed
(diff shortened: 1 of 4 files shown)


schuylermartin45:
I probably won't get to this until tomorrow, but let's at least get those CI checks running in the meantime. Thanks!

schuylermartin45 (approved):
CEP-0014 arguably should have highlighted this nuance, but this makes a lot of sense. Thanks for catching this!

I'm not sure when we'll next release CRM, but this will be in it. We always appreciate the help!

PR #542, opened 2026-07-10
Fix multiline str regression issue 541

### LINKED ISSUE #541: Multi-line string issue
**What happened?**

conda-recipe-manager can't seem to be able to parse this recipe:

```yaml
about:
  description: 'ContourPy is a Python library for calculating contours of 2D quadrilateral

    grids.  It is written in C++11 and wrapped using pybind11.

    '
  dev_url: https://github.com/contourpy/contourpy
  doc_url: https://contourpy.readthedocs.io
  home: https://github.com/contourpy/contourpy
  license: BSD-3-Clause
  license_family: BSD
  license_file: LICENSE
  summary: Python library for calculating contours of 2D quadrilateral grids.
```

**To Reproduce**

```python
with open('recipe.yaml', 'r') as fh:
    conda_recipe_manager.parser.recipe_reader.RecipeReader(fh.read())
```

**Additional Details (please complete the following information):**
 - Conda Recipe Manager version: 0.10.4

leahsch:
### Description

- Fixes [Issue-541](https://github.com/conda/conda-recipe-manager/issues/541) and a few other minor things
- Adds tests

### DIFF
--- a/conda_recipe_manager/parser/recipe_reader.py
+++ b/conda_recipe_manager/parser/recipe_reader.py
@@ -33,6 +33,7 @@ from conda_recipe_manager.parser._types import (
 )
 from conda_recipe_manager.parser._utils import (
     dedupe_and_preserve_order,
+    look_ahead_next_non_blank_line_idx,
     normalize_multiline_strings,
     num_tab_spaces,
     quote_special_strings,
@@ -242,9 +243,7 @@ class RecipeReader(IsModifiable):
 
         # Addressing issue #541.
         # Look ahead past any blank lines to find the next real content line.
-        look_ahead_idx = line_idx
-        while look_ahead_idx < len(lines) and lines[look_ahead_idx].strip() == "":
-            look_ahead_idx += 1
+        look_ahead_idx = look_ahead_next_non_blank_line_idx(lines, line_idx)
 
         # If the file ends in blank lines, there's no continuation to find.
         if look_ahead_idx >= len(lines):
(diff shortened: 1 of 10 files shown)


conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -240,9 +240,19 @@ def _parse_multiline_node_raw(
         if line_idx >= len(lines):
             return line_idx, None
 
+        # Addressing issue #541.
+        # Look ahead past any blank lines to find the next real content line.
+        look_ahead_idx = line_idx
+        while look_ahead_idx < len(lines) and lines[look_ahead_idx].strip() == "":
```
schuylermartin45: I'm surprised pylint didn't bias towards `not <str>` instead of `== ""`.

schuylermartin45: I'm not sure how much I care in this case. It reads fine either way.

conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -297,9 +307,19 @@ def _parse_multiline_node_raw_list(
         if line_idx >= len(lines):
             return line_idx
 
+        # Addressing issue #541.
+        # Look ahead past any blank lines to find the next real content line.
+        look_ahead_idx = line_idx
+        while look_ahead_idx < len(lines) and lines[look_ahead_idx].strip() == "":
```
schuylermartin45: If we do this multiple times, we should either see if we can avoid duplication or put it in a helper function.

schuylermartin45 (commented):
A few quick ideas.

conda_recipe_manager/parser/recipe_reader.py
leahsch: I was trying to be consistent with what [`_accumulate_multiline_str()`](https://github.com/conda/conda-recipe-manager/blob/main/conda_recipe_manager/parser/recipe_reader.py#L210) does to check for blank lines so we don't have an issue with them checking different things

conda_recipe_manager/parser/recipe_reader.py
leahsch: I didn't clock at first that the fix was needed twice (once in each of the two different functions) and then I debated the helper but decided to just throw it out here first
I'll move it to a helper

schuylermartin45 (approved):

PR #552, opened 2026-08-26
Improves the error messaging around duplicate keys

schuylermartin45:
For a better `crm convert` user experience. We now attempt to log a path while building the parse tree so the user can more easily find where a duplicate key is.


Callek (commented):
@schuylermartin45 I'm not sure this is exactly right ("ignoring...")  I agree this is existing behavior so I'm approving, but let me say why. 

In yaml a duplicate key takes the later-appearing key, not the first.

Callek (approved):

markan (approved):
Not an expert but seems reasonable from context.

PR #554, opened 2026-08-31
fix: ensure jinja templates are not replaced when fixing ambiguous variables

### LINKED ISSUE #307: Error converting a recipe with a set `python_min` variable
The bug was originally reported at https://github.com/hadim/feedrattler/issues/4.

When trying to convert https://github.com/conda-forge/pyhf-feedstock/blob/5f1ecaec210f588db912f57823a3d142a7c50da4/recipe/meta.yaml with `pixi exec conda-recipe-manager convert meta.yaml`, the command generates:

```yaml
schema_version: 1

context:
...
  python_min: 3.7

...

requirements:
  host:
    - python 3.7.*
```

instead of 

```yaml
schema_version: 1

context:
...
  python_min: 3.7

...

requirements:
  host:
    - python ${{ python_min }}.*
  
```

beckermr:
### Description

This PR fixes a bug where jinja2 variables are replaced while ambiguous variables are handled when converting to v1.

closes #307

### DIFF
--- a/conda_recipe_manager/parser/dependency.py
+++ b/conda_recipe_manager/parser/dependency.py
@@ -133,6 +133,11 @@ class DependencyVariable:
         # TODO normalize common JINJA functions for quote usage
         self.name = s.strip()
 
+    # This property allows the object to be used as a MatchSpec without type guards similarly to `name` above.
+    @property
+    def original_spec_str(self) -> str:
+        return self.name
+
     def __eq__(self, o: object) -> bool:
         """
         Checks to see if two objects are equivalent.
--- a/conda_recipe_manager/parser/dependency.py
+++ b/conda_recipe_manager/parser/dependency.py
@@ -133,9 +133,13 @@ class DependencyVariable:
         # TODO normalize common JINJA functions for quote usage
         self.name = s.strip()
 
-    # This property allows the object to be used as a MatchSpec without type guards similarly to `name` above.
     @property
     def original_spec_str(self) -> str:
+        """
+        Property that allows the object to be used as a `MatchSpec` without type guards similarly to `name` above.
+        
+        :returns: A string representing the original specification.
+        """
         return self.name
 
     def __eq__(self, o: object) -> bool:
--- a/conda_recipe_manager/parser/dependency.py
+++ b/conda_recipe_manager/parser/dependency.py
@@ -137,7 +137,7 @@ class DependencyVariable:
     def original_spec_str(self) -> str:
         """
         Property that allows the object to be used as a `MatchSpec` without type guards similarly to `name` above.
-        
+
         :returns: A string representing the original specification.
         """
         return self.name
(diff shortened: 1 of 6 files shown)


conda_recipe_manager/parser/dependency.py
```diff
@@ -133,6 +133,11 @@ def __init__(self, s: str):
         # TODO normalize common JINJA functions for quote usage
         self.name = s.strip()
 
+    # This property allows the object to be used as a MatchSpec without type guards similarly to `name` above.
+    @property
+    def original_spec_str(self) -> str:
+        return self.name
```
schuylermartin45:
```suggestion
    @property
    def original_spec_str(self) -> str:
        """
        Property that allows the object to be used as a `MatchSpec` without type guards similarly to `name` above.
        
        :returns: A string representing the original specification.
        """
        return self.name
```
Just moving the comment to a docstring and matching the general CRM-docstring style for consistency.

I don't personally like the `@property` decorator (I'll spare everyone that rant lol), but this is a drop-in [replacement](https://github.com/conda/conda/blob/main/conda/models/match_spec.py#L373) for the original, so we're stuck with it.

conda_recipe_manager/parser/recipe_reader_deps.py
```diff
@@ -166,12 +166,15 @@ def _extract_test_requirements(
                 )
             )
 
-    def get_all_dependencies(self, include_test_dependencies: bool = False) -> DependencyMap:
+    def get_all_dependencies(self, include_test_dependencies: bool = False, sub_vars: bool = True) -> DependencyMap:
```
schuylermartin45, line 169: This makes sense to expose, thanks for expanding the function and for defaulting to the original behavior!

conda_recipe_manager/parser/recipe_parser_convert.py
```diff
@@ -269,7 +280,7 @@ def _upgrade_ambiguous_deps(self) -> None:
                 if dep.data.version is None or not isinstance(dep.data.original_spec_str, str):  # type: ignore[misc]
                     continue
 
-                spec_str = dep.data.original_spec_str
+                spec_str = cast(str, dep_with_vars.data.original_spec_str)
```
schuylermartin45, line 283: One could argue that calling `str()` over `cast()` is ever-so-slightly-safer. But I think in this circumstance the risk is low.

I tend to use `str()` when serializing/deserializing from an external/non-typed source of information. But again, this should be pretty safe.

schuylermartin45 (commented):
@beckermr This looks really straight-foward and a great improvement overall!

I'll approve this when the tests pass and if you are okay taking in my docstring suggestion.

I was already hoping to release CRM v0.10.6 this week with some other bug fixes so I'm hoping to have a quick turnaround on this.

schuylermartin45 (approved):

PR #555, opened 2026-08-31
Pins `click` in `pyproject.toml` and upgrades `click` version

schuylermartin45:
- Pins `click` in the `pyproject.toml` file, as discovered in https://github.com/conda-forge/conda-recipe-manager-feedstock/pull/44
  - We continue to be bad at keeping the `pyproject.toml` aligned with the conda recipe file for this project.
- While we are at it, also bumps the `click` pinning. In my brief testing, this looks safe. I'll be curious to see the integration tests pass (which calls a number of CRM `click`-based scripts)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ Note: version releases in the 0.x.y range may introduce breaking changes.
 ## [Unreleased]
 ### Added
 ### Changed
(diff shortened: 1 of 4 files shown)


Callek (approved):
As I said in an Anaconda internal message. I'm not really a fan of the hard pins like this, however I'm also a big +1 on making sure things are consistent and improving it deliberately.

[I don't think my approval actually helps you on the PR, but you have it anyway]

PR #556, opened 2026-08-31
Adds package builder warning

schuylermartin45:
- Warns package builders when a V1 recipe contains "dual-context" `/requirements` sections. This is an incredibly nuanced feature of V0 that is no longer supported in V1.
- Adds a new `ROOT_PATH` constant that has an identical value as `ROOT_NODE_VALUE` but is used in a completely different meaning. This eliminates a few magic string instances of "/", in favor of said constant.
- Updates tests


Callek (approved):
This looks like it does what the PR comment says.

Issue #561, opened 2026-09-03
CRM bugs/limitations worked around in ``conda-forge-bot``'s proposed v0 -> v1 recipe migrator


pb01ka:
Quoting part of [@beckermr's comment](https://github.com/conda-forge/conda-forge-bot/pull/6381#pullrequestreview-5023295434):

> Much of the extra code here is workarounds for bugs in CRM itself. Fixes to CRM need to be made upstream along with better testing.

The following is a list of functions I implemented in my PR - https://github.com/conda-forge/conda-forge-bot/pull/6381. I have also explained the workaround each function covers. My intent is to implement fixes for these here in CRM itself (as per the suggestion of @beckermr), and once they land, remove the corresponding workaround code from my ``conda-forge-bot`` PR.

**Disclaimer:**

- It's been about a month since I added these workarounds. If CRM has since gained the capability (or if it already had then please point out) to handle any of the cases listed below, there's no need to address that item here.
- I'm leaving it up to the CRM team to decide whether any of these are worth adding as a mainstream addition to CRM. I'm just sharing the full list of pre/post-processing functions I had to add to get the R recipes (taken from R feedstocks) I worked with to migrate cleanly. A copy of these R recipes is available in [pb01ka/r-feedstock-v1-feasibility](https://github.com/pb01ka/r-feedstock-v1-feasibility).

#### Coming from `conda_forge_tick/migrators/v0_to_v1.py` in my [`r-migrator` branch](https://github.com/pb01ka/conda-forge-bot/tree/r-migrator):

- [ ] [`_duplicate_top_level_key_reasons`](https://github.com/pb01ka/conda-forge-bot/blob/978a5ac16368365b80158123781204724f611647/conda_forge_tick/migrators/v0_to_v1.py#L62): CRM silently accepts two unconditional top-level sections with the same name (e.g. two `build:` blocks) via `ALLOW_DUPLICATE_KEYS` instead of rejecting them, and the duplicate survives into the v1 output where it's only caught later as an opaque `DuplicateKeyException`. This function detects the pattern in the raw v0 text first and reports it with a clear, actionable reason.
[...]

PR #558, opened 2026-09-02
fix: handle more ambiguous spec edge cases for simple jinja2 statements and in test deps

beckermr:
### Description

This PR fixes edge cases where the conversion function does not add `.*` to specs like `python {{ python_min }}` when `python_min` does not have a corresponding set statement.

It also fixes cases where ambiguous test deps were not converted.

closes #307

PR #564, opened 2026-09-15
Reject duplicate keys that CRM cannot safely merge into V1 output

pb01ka:
### Description

- Addresses the `_duplicate_top_level_key_reasons` workaround listed in https://github.com/conda/conda-recipe-manager/issues/561 (from `conda-forge-bot`'s v0 -> v1 migrator), by fixing the underlying gap in CRM itself.
- `ALLOW_DUPLICATE_KEYS` previously tolerated any duplicate key, silently carrying invalid duplicates into V1 output (e.g. two full `build:` blocks, or a selector-guarded `run:` list block).
- As far as I could tell, CRM only has a real fold for a duplicated leaf key (`script: ... # [win]`) when every occurrence carries a selector - it becomes a valid V1 ternary. Everything else (duplicated sections, or a leaf missing a selector on even one occurrence) is carried through unmerged, producing malformed V1.
- Adds `RecipeReader._get_invalid_key_duplication_reason()`, called on every duplicate key regardless of `ALLOW_DUPLICATE_KEYS`:
  - Rejects a duplicated section (a key introducing nested content) outright - no selector can make it safely mergeable.
  - Rejects a duplicated leaf key unless every occurrence carries a selector.
- `DuplicateKeyException` add a `reason` field, so the raised message explains why the duplicate is invalid, not just where it is.
- Adds test fixtures/cases: unconditional duplicate section, selector-guarded duplicate section, nested duplicate section, leaf key missing a selector on one occurrence, and a unit test for the `_is_leaf_key()` single-list-item edge case.

#### Generative AI Disclosure

> Since, [as per this review comment](https://github.com/conda/conda-recipe-manager/pull/564#pullrequestreview-5249159088) no AI policy is in effect (as of making this edit) for this project. Therefore, taking into account my personal opinions on AI usage, I consider that adding this section is one of many ways to declare AI usage in this PR.

- Model used: [Claude Sonnet 5](https://www.anthropic.com/news/claude-sonnet-5)
- The purposes for which the said model was used:
  - Brainstorming different solution ideas that I had for the problem explained above
  - Adding/Refining docstrings/code-comments in the functions I implemented in this PR
  - For self-reviewing the changes in this PR
- Before pushing, all the changes made in the commits of this PR are reviewed by me thoroughly.

Please let me know if you want me to declare my AI usage in a different manner.


pb01ka:
@schuylermartin45 @beckermr Please take a look at this PR. TY.

pb01ka:
Quoting the failure reasoning from https://github.com/conda/conda-recipe-manager/actions/runs/34987481554/job/104468414744?pr=564

> conda 26.7.2 has requirement ruamel-yaml<0.19,>=0.11.14, but you have ruamel-yaml 0.19.1.

This seems unrelated to my changes. I think I can send a separate PR for fixing this. What do you say @beckermr @schuylermartin45?

jaimergp:
> conda 26.7.2 has requirement ruamel-yaml<0.19,>=0.11.14, but you have ruamel-yaml 0.19.1.

This will be addressed in conda 26.9, no biggie. Already in `main`.

schuylermartin45:
Apologies for the delay on this, I didn't have much time yesterday to get to this email.

We appear to have hit the same problem in an unrelated issue with our conda-forge feedstock: https://github.com/conda-forge/conda-recipe-manager-feedstock/pull/47

I will try to get to reviewing this PR today but it may have to wait until Thursday or Friday. Thanks!

conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -767,6 +767,58 @@ def _init_schema_version_and_sanitize_v0_yaml(
 
         return str(sanitized_fmt), tof_comment_cntr
 
+    @staticmethod
+    def _is_leaf_key(node: Node) -> bool:
```
schuylermartin45: This should be a member in the `Node` class. That would simplify things and better separate concerns.

conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -767,6 +767,58 @@ def _init_schema_version_and_sanitize_v0_yaml(
...
+        section that introduces nested content on subsequent lines.
+
+        NOTE: `Node.is_single_key()` is not sufficient here: a section holding exactly one list item (e.g.
+        `run:` followed by a single line) also ends up with exactly one "strong leaf" child, indistinguishable
+        from a true in-line pair. The child's `list_member_flag` disambiguates the two.
+
+        :param node: The key node to inspect.
+        :returns: True if the node's value was declared on the key's own line. False otherwise.
+        """
+        return node.is_single_key() and not node.children[0].list_member_flag
```
schuylermartin45: `is_single_key()` internally checks for a list containing at least 1 member. So I suppose this is safe but some part of me thinks we shouldn't rely on the current implementation to make that assumption for us.

I guess I'm being a little pedantic on OOP best practices on encapsulation. Again, moving this function to be a part of the `Node` class would alleviate this concern.

conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -767,6 +767,58 @@ def _init_schema_version_and_sanitize_v0_yaml(
...
+
+        Additionally, even a duplicated leaf key only folds correctly when every occurrence carries a selector
+        (e.g. `# [win]`): without one on every occurrence, there is no reliable way to tell which value is the
+        fallback/default, and the V1 fold itself produces broken output.
+
+        :param new_node: The newly parsed, duplicate key node.
+        :param parent: The parent node both the new node and its existing duplicate(s) live under.
+        :returns: `None` if the duplication can be tolerated. Otherwise, a reason to explain why it cannot.
+        """
+        occurrences = [child for child in parent.children if child.value == new_node.value] + [new_node]
```
schuylermartin45:
```suggestion
        occurrences: Final = [child for child in parent.children if child.value == new_node.value] + [new_node]
```

conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -767,6 +767,58 @@ def _init_schema_version_and_sanitize_v0_yaml(
...
+
+        if not all(RecipeReader._is_leaf_key(occurrence) for occurrence in occurrences):
+            return (
+                "This key introduces a section (nested content on following lines), and a duplicated section "
+                "cannot be safely merged - there is no way to reconcile two blocks of nested content into one. "
+                "This is likely an authoring mistake in the source recipe that needs manual review."
```
schuylermartin45: This is incredibly wordy and I think should be simplified. The error message will not easily fit in a terminal window for some users.

conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -767,6 +767,58 @@ def _init_schema_version_and_sanitize_v0_yaml(
...
+            )
+
+        if not all(Regex.SELECTOR.search(occurrence.comment) for occurrence in occurrences):
+            return (
+                "Every occurrence of a duplicated key must carry a selector (e.g. `# [win]`) to distinguish it "
```
schuylermartin45: Again, this should be simplified. I believe the term `selector` is widely understood enough in the V0 format by the Conda Community that we don't need to give an example.

conda_recipe_manager/parser/recipe_reader.py
```diff
@@ -858,6 +910,10 @@ def _construct_parse_tree(self, sanitized_yaml: str, tof_comment_cntr: int) -> N
                 and not new_node.list_member_flag
                 and new_node.value in [child.value for child in parent.children]
             ):
+                invalid_dup_reason = RecipeReader._get_invalid_key_duplication_reason(new_node, parent)
```
schuylermartin45, line 910:
```suggestion
                invalid_dup_reason: Final = RecipeReader._get_invalid_key_duplication_reason(new_node, parent)
```

tests/parser/test_recipe_reader.py
```diff
@@ -1879,3 +1880,119 @@ def test_duplicate_keys_allowed(file: str, flags: RecipeReaderFlags, caplog: pyt
...
+
+    :param file: File to run against
+    :param flags: RecipeReaderFlags to enable.
+    """
+    with pytest.raises(DuplicateKeyException) as e:
+        load_recipe(file, RecipeReader, flags)
+    assert e.value.message == (
+        "Duplicate key found at line 13: build. This key introduces a section (nested content on following lines), "
+        "and a duplicated section cannot be safely merged - there is no way to reconcile two blocks of nested "
+        "content into one. This is likely an authoring mistake in the source recipe that needs manual review."
```
schuylermartin45: To keep things consistent with the rest of the project and to make it easier to add future test cases, I think these test cases should be combined and this `assert` should be comparing against an `expected` message value.

Maybe the messages should be saved in a `Final`-constant so they are easier to compare against if more test cases are added in the future.

Also doesn't `e` have to be within the `with` block to be valid? Won't `e` "go out of scope" as soon as the `with` block terminates? At the very least I think it reads better.

tests/parser/test_recipe_reader.py
```diff
@@ -1879,3 +1880,119 @@ def test_duplicate_keys_allowed(file: str, flags: RecipeReaderFlags, caplog: pyt
...
+    :param file: File to run against
+    :param flags: RecipeReaderFlags to enable.
+    """
+    with pytest.raises(DuplicateKeyException) as e:
+        load_recipe(file, RecipeReader, flags)
+    assert e.value.message == (
+        "Duplicate key found at line 11: script. Every occurrence of a duplicated key must carry a selector (e.g. "
+        "`# [win]`) to distinguish it from the others; otherwise there is no reliable way to tell which value "
+        "should apply."
+    )
```
schuylermartin45: Yeah this is a lot of repeated boiler-plate work that should be condensed by using an extra parameterized `expected` argument.

tests/parser/test_recipe_reader.py
```diff
@@ -1879,3 +1880,119 @@ def test_duplicate_keys_allowed(file: str, flags: RecipeReaderFlags, caplog: pyt
...
+
+
+def test_is_leaf_key_rejects_section_with_single_list_item() -> None:
```
schuylermartin45: This should be moved to the tests for `_node.py` if we are moving the function. That also means we don't have to import `Node` into this test, further separating concerns.

schuylermartin45 (commented):
I'm going to point out that this change negatively impacts the current integration test results. We will be increasing our exception count from 75 recipes to 126.

To my knowledge, this is the first time we've knowingly had an increase over a change. I get that this is desired behavior in conda-forge's purposes but I want to have a think on this before we make the change universal. We need to ensure (as best as we can) that changes work for all users of CRM in the Conda Community.

I think this is probably fine, but I'm open to discuss it further. At the very least we should probably expand the documentation on `ALLOW_DUPLICATE_KEYS` to capture the nuance introduced. The original intent was to support V0 selector'd duplicate keys, not realizing that non-selector'd V0 duplicate keys were a common problem.

Additionally if generative-AI was used to create this work, I would like to see that disclosed somewhere. I will admit that there is no current policy for this project, but we are in the midst of creating one, likely modeled after the conda project, but amended to require disclosure of which tools were used.

I see this as an important ask to keep Open Source Software decision making transparent and available to the wider community. Certain decisions and the way the docstrings are written make me believe generative-AI was used.

beckermr:
> I'm going to point out that this change negatively impacts the current integration test results. We will be increasing our exception count from 75 recipes to 126.

@schuylermartin45 were those recipes converted correctly and are now falsely being reported as uncovertable, or were those recipes converted incorrectly before and now the reporting is more accurate?

I think if it is the second thing, then the change is good and it is fixing a bug.

schuylermartin45:
> > I'm going to point out that this change negatively impacts the current integration test results. We will be increasing our exception count from 75 recipes to 126.
> 
> @schuylermartin45 were those recipes converted correctly and are now falsely being reported as uncovertable, or were those recipes converted incorrectly before and now the reporting is more accurate?
> 
> I think if it is the second thing, then the change is good and it is fixing a bug.

The stats in the summary review aren't that detailed. Presumably they were getting converted but with the invalid duplicate key, which I know is the point of this PR.

I know we've spoken and conda-forge would like these to error but I could see other users preferring to get a partially correct conversion that they may have to manually edit.

But I guess we are also providing explicit reasons _why_ we threw which would also be easy to debug manually. Maybe I'm blowing this out of proportion for a relatively small set of recipe files.

beckermr:
> The stats in the summary review aren't that detailed. Presumably they were getting converted but with the invalid duplicate key, which I know is the point of this PR.
>
> I know we've spoken and conda-forge would like these to error but I could see other users preferring to get a partially correct conversion that they may have to manually edit.
>
> But I guess we are also providing explicit reasons why we threw which would also be easy to debug manually. Maybe I'm blowing this out of proportion for a relatively small set of recipe files.

These are all fair points. I think per previous discussions we'd be happy with a flag where we can have CRM fail for some errors. This would be a lot like the compiler flags where you can do `-Wno-...` and could include erroring for all warnings (though we would not use that mode).

Right now there is only one such flag, and its name is ad hoc. I think if we go this route, we should devise a more general system. I expect there to be more than 2 cases like this.

schuylermartin45:
> Right now there is only one such flag, and its name is ad hoc. I think if we go this route, we should devise a more general system. I expect there to be more than 2 cases like this.

I'd agree; let's make a general flag. As much as I like the compiler flags, I think for our purposes I'm okay if we have a simpler all-or-nothing flag. At least to start, and if we have to get granular we can revisit/iterate.

conda_recipe_manager/parser/recipe_reader.py
pb01ka: Upon thinking from both perspectives (i.e., in favour of and against the idea of making `_is_leaf_key` a member of the `Node` class),

- A necessary condition for checking the merge-ability of a `Node` is that it should be inline. The other condition is that all of the duplicates should have selectors.
- Now, I have only one concrete argument of keeping it here which is that it serves a very specific purpose of partially figuring out whether the `node: Node` parameter is merge-able. This motivated me to add it as a `staticmethod` in `RecipeReader` - the only place where such a method would be used. I felt that it's not generic enough property to be added as a member of `Node`. I also wasn't sure, as a user of `Node` objects, whether I really want to know if a node is merge-able or not. However, a `RecipeReader` is indeed always interested in knowing the same.
- The argument to move it as a member method of `Node` class - As said above, it is indeed a property of a `Node` (no matter how specific it is). It accepts a `Node` typed parameter and performs simple checks on it.

Considering both I think I would move it to `Node` but,

- I'd rename it to `is_inline_scalar_key` so that it highlights (to some extent) the purpose it is serving.

Thanks for your opinions/thoughts.

conda_recipe_manager/parser/recipe_reader.py
pb01ka: > `is_single_key()` internally checks for a list containing at least 1 member

Please correct me if I'm misinterpreting - I see that your concern is in the fact that while doing `node.children[0]` we are assuming that `is_single_key` already performed `len(self.children) == 1` check.

If I'm right in picking up your concern, then here are some pointers,

- What if `is_single_key` removes `len(self.children) == 1`? - It will still do the same indexing (i.e., `self.children[0]`) as we are doing (i.e., `node.children[0]`). It is also note worthy that we will only reach `node.children[0]` only if `node.is_single_key()` succeeds and returns `True` (otherwise `and` would just short circuit). This would in-itself mean that our assumption will always hold when `node.children[0].list_member_flag` check executes.
- What if `is_single_key` also removes `self.children[0].is_strong_leaf()` - If this ever happens then `is_single_key` would collapse to `is_key` - basically it won't serve any purpose.
- What if `is_single_key` implements its check in an entirely different way - If this happens, even then the contract it promises in its public API - "Indicates if a node contains a single child node and is a key." - will still be satisfied. IOW, if it returns `True` then it means that the node has a single child i.e., `len(self.children) == 1` holds.
- It is also worth quoting the following snippet from `main` branch,

https://github.com/conda/conda-recipe-manager/blob/9989d58867b330a8e22b4123278190d4b7c183d0/conda_recipe_manager/parser/recipe_reader.py#L1133-L1135

---

Considering all the above pointers, my response to your comment is as follows,

> we shouldn't rely on the current implementation to make that assumption for us.

We are not relying on the implementation of `is_single_key` to assume that the condition `len(self.children) == 1` is True. Instead we are relying on the contract it communicates via its public doc string.

> I guess I'm being a little pedantic on OOP best practices on encapsulation

Continuing from my reasoning above, I don't think that my assumption is connected to encapsulation.

> Again, moving this function to be a part of the `Node` class would alleviate this concern.

As per my analysis from above - I think this is argument is not stronger than the one given [in my above comment](https://github.com/conda/conda-recipe-manager/pull/564#discussion_r4049209481).

tests/parser/test_recipe_reader.py
pb01ka: > This should be moved to the tests for `_node.py` if we are moving the function

On the basis of as far as I can see in the code base,

- There is no test file specifically for `_node.py`. The file is located at `conda_recipe_manager/parser/_node.py` (inside the `parser` directory).
- Therefore I checked the files in https://github.com/conda/conda-recipe-manager/tree/main/tests/parser
- Also, I couldn't find any file inside [`tests`](https://github.com/conda/conda-recipe-manager/tree/main/tests) directory which creates `Node` object and tests its member methods.

Considering all the above points I intend to create a new file and move this test into it.

conda_recipe_manager/parser/recipe_reader.py
pb01ka: Upon introducing `Final` with `invalid_dup_reason` I am getting the following error with `mypy` 2.3.1,

```
(py313) 1:00:13:~/.../conda-recipe-manager % python -m mypy conda_recipe_manager/parser/_node.py conda_recipe_manager/parser/recipe_reader.py 

conda_recipe_manager/parser/recipe_reader.py:525: error: Unused "type: ignore" comment  [unused-ignore]
                    yaml.dump(value, Dumper=ForceIndentDumper, sort_keys=False, width=sys.maxsize),  # type: ignore[mis...
                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
conda_recipe_manager/parser/recipe_reader.py:914: error: Cannot use Final inside a loop  [misc]
                    invalid_dup_reason: Final = RecipeReader._get_invalid_key_duplication_reason(new_node, parent)
                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Found 2 errors in 1 file (checked 2 source files)
(py313) 1:00:46:~/.../conda-recipe-manager % python -m mypy --version
mypy 2.3.1 (compiled: yes)
```

Therefore I won't be able to apply this suggestion as is.

tests/parser/test_recipe_reader.py
pb01ka: I would first address your comments/questions regarding `e`.

---

> Won't `e` "go out of scope" as soon as the `with` block terminates?

To the best of my knowledge, my answer to this question is - No, `e` doesn't go "out of scope" when the `with` block terminates. This is because `with` statement in Python doesn't create a scope. Some references to support my point,

- https://stackoverflow.com/a/45100308
- https://docs.python.org/3/reference/compound_stmts.html#index-18 (especially the 5th and the 7th steps - no scope is created or deleted).

---

> Also doesn't `e` have to be within the with block to be valid?

As per my answer above, no, `e` doesn't have to be within the block to be valid.

---

> At the very least I think it reads better.

The assertion style for `e.value.message` is similar to what's shown in https://docs.pytest.org/en/stable/how-to/assert.html#assertions-about-expected-exceptions

In addition to this, some snippets from the `main` branch,

https://github.com/conda/conda-recipe-manager/blob/9989d58867b330a8e22b4123278190d4b7c183d0/tests/parser/test_recipe_reader.py#L1849-L1851

https://github.com/conda/conda-recipe-manager/blob/9989d58867b330a8e22b4123278190d4b7c183d0/tests/parser/test_recipe_reader.py#L1872-L1874

So if we want to change the assertion style for `e.value.message`, then I'd prefer to do it in a PR specifically dedicated for this so that the update can happen everywhere uniformly. Please let me know if you think otherwise.

conda_recipe_manager/parser/recipe_reader.py
schuylermartin45: I missed that context, it wasn't shown in the diff-view.

conda_recipe_manager/parser/recipe_reader.py
schuylermartin45: I'm okay with the name, I do not think it should be a `staticmethod`.

tests/parser/test_recipe_reader.py
schuylermartin45: I'm not asking for much and I did not expect this to be a controversial suggestion. Please make the changes without refactoring the existing issues. We can address them another time.

And again, I would very much appreciate disclosure of the generative AI tooling being used here.

pb01ka: > I think these test cases should be combined and this `assert` should be comparing against an `expected` message value.

I agree with you on this.

- I intend to de-duplicate `test_duplicate_unconditional_section_always_raises`, `test_duplicate_selector_guarded_section_always_raises`, `test_duplicate_nested_section_always_raises`, and `test_duplicate_leaf_key_missing_selector_on_an_occurrence_always_raises`
- That being said I plan to [use custom test IDs](https://docs.pytest.org/en/stable/example/parametrize.html#different-options-for-test-ids) so that its easier to debug in the future.

---

> Maybe the messages should be saved in a `Final`-constant so they are easier to compare against if more test cases are added in the future.

Noted. 

Thank you for the ideas.

pb01ka: > And again, I would very much appreciate disclosure of the generative AI tooling being used here.

I just saw your latest comments (appeared only when I reloaded the page). I will update the PR description once I address/reply to all of your reviews and push the changes.

conda_recipe_manager/parser/recipe_reader.py
pb01ka: Addressed in https://github.com/conda/conda-recipe-manager/pull/564/commits/f241cfba42a628b8ff16f65bd1be93993910974c

conda_recipe_manager/parser/recipe_reader.py
pb01ka: Addressed in https://github.com/conda/conda-recipe-manager/pull/564/commits/f241cfba42a628b8ff16f65bd1be93993910974c

pb01ka:
> Right now there is only one such flag, and its name is ad hoc. 

(AFAICT), are you talking about, `--fail-on-unsupported-jinja` (see the snippet below),

https://github.com/conda/conda-recipe-manager/blob/9989d58867b330a8e22b4123278190d4b7c183d0/conda_recipe_manager/commands/convert.py#L316-L324

**If yes**:

From my understanding of [the above comment](https://github.com/conda/conda-recipe-manager/pull/564#issuecomment-5733266365),

> As much as I like the compiler flags, I think for our purposes I'm okay if we have a simpler all-or-nothing flag

Do you mean that there should be a single flag like, `--fail-on-warnings`/`--warnings-as-errors`/`--strict-mode` which would toggle between all kinds of warnings and errors?

---

**Regarding reviews** - @schuylermartin45 I have replied to and addressed all of your reviews with explanations/references. Please feel free to take a look at the latest change set whenever possible. TYSM for the reviews, it was fun addressing them. 🤝 🧐 

Also I have added **Generative AI Disclosure** section in [my PR description](https://github.com/conda/conda-recipe-manager/pull/564#issue-5462371155).

tests/parser/test_recipe_reader.py
pb01ka: > Please make the changes without refactoring the existing issues

Addressed in https://github.com/conda/conda-recipe-manager/pull/564/commits/0bb3f6f31804dcc8ef3dfed58abfab2f3fcaf944

beckermr:
So failing on all warnings won't work since CRM warns about some things that  are warnings but not fatal.

We want a flag that causes CRM to fail if the recipe it produces is known to be incorrect or incomplete. I use this function to post-process CRM warnings to figure out which ones to ignore and which ones to let pass through:

```python
def _post_process_returncode_and_stderr(returncode, stderr):
    deprecated_fields = [
        # about
        "prelink_message",
        "license_family",
        "identifiers",
        "tags",
        "keywords",
        "doc_source_url",
        "license_family",
        # build
        # we do not ignore these since
        # folks need to do something with them.
        # "pre-link",
        # "noarch_python",
        # "features",
        # "msvc_compiler",
        # "requires_features",
        # "provides_features",
        # "preferred_env",
        # "preferred_env_executable_paths",
        # "disable_pip",
        # "pin_depends",
        # "overlinking_ignore_patterns",
        # "rpaths_patcher",
        # "post-link",
        # "pre-unlink",
        # "pre-link",
    ]
    new_lines = []
    for line in stderr.splitlines():
        line = line.strip()
        # ignore field deprecations
        if any(
            f"/{df}" in line and "Field at" in line and "no longer supported"
            for df in deprecated_fields
        ):
            continue


        if (
            "Recipe upgrades cannot currently upgrade ambiguous version constraints on dependencies that use variables"
            in line
        ):
            continue


        if "The recipe parser was unable to evaluate the JINJA expression" in line:
            continue


        if "Could not patch unrecognized license" in line:
            continue


        if "Version on dependency changed to: python {{ python_min }}.*" in line:
            continue


        if (
            " error" in line
            and " and " in line
            and " warning" in line
            and line.startswith("0")
        ):
            continue


        new_lines.append(line.strip())


    if not new_lines:
        return 0, stderr
    else:
        return returncode, stderr
```

The bits here are:

- There are some fields which are deprecated and removed in v1 that we can ignore. Others indicate something someone has to think about. So I only let it fail on warnings for some of them.
- Some of the warnings, like `Version on dependency changed to: python {{ python_min }}.*`, probably should not be emitted at all. In this specific case, v1 recipes forbid versions without some sort of operator (e.g., `20` vs `20.*`), and v0 recipes with versions like `20` always effectively convert to `20.*`. So there is no reason to emit the warning.
- The license warning `Could not patch unrecognized license` is another case where it is informative, but we can ignore it. We were ok with an unrecognized license at the start, and so should be fine with it now.
- The "Recipe upgrades cannot currently upgrade ambiguous version constraints on dependencies that use variables" warning is a tricky one. We ignore it, but it does need better handling. See https://github.com/conda/conda-recipe-manager/pull/558 where I had to go the route of putting in a custom pre-processing step for conda-forge before conversion, which is exactly what we are trying to prevent here. Sigh.
- The warning "The recipe parser was unable to evaluate the JINJA expression" is another case of information that is not fatal. Variables referenced in jinja can be missing and so the expression can fail to render. That is a non-issue. The only issue that needs a warning that causes a failure is if the jinja cannot be converted to acceptable v1 syntax (e.g., apparently doesn't allow some of the `.split()`-type manipulations and instead requires filter like `| split()`)
- We currently do allow failures if CRM cannot relocate comments in the file.

So the goal isn't to fail on anything weird, but to fail when the output recipe is knowingly wrong, not functionally equivalent to the original, or definitely won't parse/build due to invalid syntax.

Given CRM's current approach of warning on a lot of things, not all of which  are fatal, a single "error for all warnings" flag won't work. This is why I suggested a more general system of flags to control when CRM errors versus not.

If that solution is not acceptable, then I'd suggest a different approach:

- We dispense with all of the extra flags.
- Have CRM never fail in the sense of never returning a non-zero exit code.
- Devise a categorization of the various warnings with known codes so that others do not have to rely on matching specific error strings to understand what happened.
- Have CRM output all of the warnings to stderr in some parseable format (e.g., JSON or a simple text table).
- Defer all decisions of what is fatal versus not to the user.

beckermr:
To pile on a bit, this specific issue is around "Reject duplicate keys that CRM cannot safely merge into V1 output". Namely, the produced recipes are knowingly wrong/bad. I get the idea that someone would want to go through the output and fix things later, but I do not get why CRM would want an exit code of 0 for a recipe it knows it cannot convert correctly. Those things are not in conflict. CRM can still output what it has to stdout while also failing with a non-zero exit code.

pb01ka:
> So the goal isn't to fail on anything weird, but to fail when the output recipe is knowingly wrong, not functionally equivalent to the original, or definitely won't parse/build due to invalid syntax.

I agree. I also think that's what we are doing here. We are handling the case of duplicate keys that cannot be merged/reconciled safely which, I think falls in "definitely won't parse/build due to invalid syntax" category.

---

> I do not get why CRM would want an exit code of 0 for a recipe it knows it cannot convert correctly.

> CRM can still output what it has to stdout while also failing with a non-zero exit code.

I agree with both of your points. I think this approach has merits. That being said,

- Theoretically if the input v0 recipe has keys which cannot be merged, then it cannot be **meaningfully**  transformed to v1 format. This is because the input recipe is semantically incorrect.
- However, the same input recipe can be converted into v1 format by letting both the duplicate keys through as is i.e., by not attempting to merge them (the current status in `main`). It won't be a meaningful transformation but it will still produce some output which the users can edit.
- **Overall**: I'd say, rather than rejecting the recipe outright (which is what this PR is doing), we could keep letting duplicate keys through (while skipping any merge attempt), and return a non-zero exit code with the same error/reason we already compute here.

What do you say?

---

> If that solution is not acceptable, then I'd suggest a different approach:

This is better approach out of the two, you suggested. This allows for catering to the needs of all the users while still flagging theoretically incorrect inputs/outputs.

PR #565, opened 2026-09-16
fix parser crashing when a comment is indented deeper than its siblings

leahsch:
### Description
Fixes a parsing issue with comment lines that were indented deeper than sibling lines in the same section:
```
requirements:
  run:
    - foo
     # a stray comment indented one space too many
     # another one
    - bar
```
This was crashing with the error: `AttributeError: 'list' object has no attribute 'setdefault'` because the comment line was being read as the start of a new indentation level.

Added a guard to handle an extra indented comment line as a sibling to the previous node unless the previous node is a key with no new children, _yet_.

Added tests.


schuylermartin45 (approved):
This looks pretty straight forward and fixes 5 recipes in the integration tests that would otherwise have thrown.

Previous Run: https://github.com/conda/conda-recipe-manager/actions/runs/34239764522
This Run: https://github.com/conda/conda-recipe-manager/actions/runs/35149962827?pr=565

schuylermartin45:
The build issue is a known dependency problem with the current version of the `conda` package on conda-forge. I will work my magic to get this in, in the meantime.

schuylermartin45 (approved):

